PR #4, opened 2024-08-08
Improving sessions logic and start conversation button

fabclj:
# Success criteria

- This PR improves the behavior of the Start Conversation buttons, in Home Screen. The goal is to prevent the creation of new session on every button click. If a session is already active, it opens to that session. 
- It was refactored the logic regarding the socket connection, in case of Home Screen active.
- We now connect to the socket only if the user hit the Start Conversation button from home or from teaser messages or from Previous conversation.
- It fixes a bug where Conversation starters button in Home Page and Teaser message, in case are disabled and some button is configured, it displays anyway on UI.

# How to test
Test a new endpoint with all different configurations, with and without persistent userId / sessionId
1. Basic config without Home Screen, Privacy Screen or teaser message.
2. Config with Homepage active and/or Privacy Screen active
3. With action button in Home screen and/or in Teaser message and preview messages

### DIFF
--- a/cypress/e2e/engagement.cy.ts
+++ b/cypress/e2e/engagement.cy.ts
@@ -79,6 +79,9 @@ describe("Engagement Message", () => {
 				unreadMessages: {
 					enablePreview: false,
 				},
+				homeScreen: {
+					enabled: false,
+			}
 			},
 		});
 		cy.get("[data-cognigy-webchat-toggle]").click().click();
@@ -137,6 +140,9 @@ describe("Engagement Message", () => {
 				unreadMessages: {
 					enablePreview: false,
 				},
+				homeScreen: {
+					enabled: false,
+			}
 			},
 		});
 		cy.wait(500);
--- a/src/common/interfaces/message.ts
+++ b/src/common/interfaces/message.ts
@@ -6,7 +6,7 @@ export interface IBaseMessage {
 	prevMessage?: IMessage;
 	source: string;
 	text?: string;
-	timestamp: number;
+	timestamp?: number;
 }
 
 export interface IUserMessage extends IBaseMessage {
--- a/src/webchat-ui/components/WebchatUI.tsx
+++ b/src/webchat-ui/components/WebchatUI.tsx
@@ -741,6 +741,26 @@ export class WebchatUI extends React.PureComponent<
 				options,
 			});
 		} else {
+			this.props.onSwitchSession();
+			this.props.onSendMessage(text, data, options);
+		}
+	};
+
+	handleSendActionButtonMessageFromTeaser = (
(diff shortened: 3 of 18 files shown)


sushmi21:
I see conversation starters in teaser message, if conversation starters are disabled but the starter buttons are configured.

sushmi21:
when I come to home screen from the middle of a conversation, I could still click the starter button in home screen and continue the current conversation.
Screenshot from 2024-08-13 16-33-56

fabclj:
> when I come to home screen from the middle of a conversation, I could still click the starter button in home screen and continue the current conversation. Screenshot from 2024-08-13 16-33-56

Probably the best behavior in this case is to always create a new session/conversation when hit the starters buttons form Homescreen. I will ask the PO for confirmation.

fabclj:
> I am confused on what configuration should have prio (the settings in index.html or the settings from endpoint editor?).
> 
> * If I disable conversation starters in index.html and enable them in endpoint settings, the starter buttons are enabled.
> * If I have one button configured in index.html (**Button 1**) and two buttons configured in endpoint settings (**Button A, Button B**) then in the UI I see **Button 1**, **Button B**

These issues about Conversation starters are unrelated to the sessions, but I agree on fix in this PR, as I did for the Homescreen starters.
I will ask our PO about this kind of priority, for sure we need to fix that.

kwinto (approved):
lgtm

kwinto:
> > I am confused on what configuration should have prio (the settings in index.html or the settings from endpoint editor?).
> > 
> > * If I disable conversation starters in index.html and enable them in endpoint settings, the starter buttons are enabled.
> > * If I have one button configured in index.html (**Button 1**) and two buttons configured in endpoint settings (**Button A, Button B**) then in the UI I see **Button 1**, **Button B**
> 
> These issues about Conversation starters are unrelated to the sessions, but I agree on fix in this PR, as I did for the Homescreen starters. I will ask our PO about this kind of priority, for sure we need to fix that.
> 
> **EDIT:** I had a look on our reducers and I realized we are doing a deep merge in this case.
> 
> https://github.com/Cognigy/Webchat/blob/ecfc550a58dc2a2c9baf094b3ec69785890f61bc/src/webchat/store/config/config-reducer.ts#L240
> 
> In order to fix that, before to perform a deep merge, we should first parse the two objects and if the buttons are disabled, we make the array empty. Also, we need to define a priority in case both are enabled and include the buttons from the prio object only. I believe this is something should be handled in a specific PR and discuss with PO.

I would not fix this in this PR though, as is out-of-scope. Maybe we can track this a fix it later.

sushmi21 (approved):

PR #39, opened 2024-11-15
Bug/80600 History restore issues

kwinto:
This PR fixes an issue where calling webchat.open() would not properly restore the history. We fixed it by ensuring the same logic applies here as in the "Start conversation" handler. 

This cries for refactoring. I've added some comments to the relevant parts for the future endeavours.  

Besides, we also added a new analytics event which enables the customers to subscribe to session switch event.
It should be helpfull for storing the updated session Id when switching conversations. The docs were updated accordingly (please test the code example!)

To test, use static sessionId and webchat.open()


sushmi21:
The session restoration is working as expected. But I read the reproduction steps(see below steps) in the bug ticket. I could still reproduce the behavior mentioned by the customer. 
    1. Set up persistent history
    2. Open Webchat and start a conversation.
    3. Refresh the page— the conversation persists as expected.
    4. Switch to a different conversation using the conversation history feature.
    5. Refresh the page again— the initial conversation from step 3 reopens, instead of the most recently active one (from step 4).

But the current behavior seems logical, although it is not what the customer expects. Is this PR intended to fix the customer's issue?

kwinto:
> But the current behavior seems logical, although it is not what the customer expects. Is this PR intended to fix the customer's issue?

If we switch the conversation, the session ID is updated. It is not the one we had init the Webchat with. 
For that, I've introduced the new analytics event. 

So we need to:
1. Store the ID when session is changed in the Browser Storage.
2. Use it when initializing Webchat.

I have updated the persistent feature documentation. But I had no time to test the code snippet. Please test it.

sushmi21:
> I have updated the persistent feature documentation. But I had no time to test the code snippet. Please test it.

This works as expected. Thanks!

sushmi21 (dismissed):

sushmi21:
A test related to engagement message display is failing. I could also reproduce the functionality of the failing test in the webchat. I will investigate this further and see if this is a regression caused by the changes in this PR or a bug. Unfortunately, we need to wait for the merge and the webchat release

sushmi21 (dismissed):

sushmi21 (approved):

PR #51, opened 2025-01-08
Make QRs in chat options post to the active session

kwinto:
This PR fixes an issue where Quick Replies from the Chat Options Screen would post to a new session insead of the active one.

To test, configure QRs in Chat Options Endpoint settings.


sushmi21 (approved):
Works as xpected

PR #55, opened 2025-01-20
Feature/83694 streaming markdown handling

dshire:
# Success criteria

This PR adds support for markdown rendering and streaming messages (animation + message chunk collation) to the chat components. There is a new view on the demo page as well.

- Markdown rendering can be used without the cognigy PR, but might have issues, please only test with the related cognigy repo PR (issues would be due to some trim() logic in cognigy that was happening before, which breaks valid markdown in some situations)

# How to test

For testing, checkout the chat-components branch of the same name with the PR linked here: https://github.com/Cognigy/chat-components/pull/106
`npm ci && npm pack` in chat-components repo,
then `npm i ../chat-components/cognigy-chat-components-0.38.0.tgz && npm run dev` in (this) Webchat repo (correct path if it is somewhere else for you)

In the local Webchat index.html you need to enable
behavior: {
collateStreamedOutputs: true,
progressiveStreaming: true,
renderMarkdown: false,
}
or test the settings in the Endpoint settings

Please test directly with the related PR in cognigy AI (coming up)

# Security
- [ ] XSS

# Additional considerations

### DIFF
--- a/docs/css-customization.md
+++ b/docs/css-customization.md
@@ -16,6 +16,7 @@ There are several classes that you need to take in consideration if you want to 
 * *webchat-header-close-button*
 * *webchat-header-rating-button*
 * *webchat-chat-history*
+* *webchat-scroll-to-bottom-button*
 * *webchat-message-row*
 * *regular-message*
 * *webchat-avatar*
@@ -140,6 +141,13 @@ This is the element where all the messages of the chat are are shown, you could 
 [data-cognigy-webchat-root] [data-cognigy-webchat].webchat .webchat-chat-history {
 
     background-color: rgb(48, 48, 48);
+}
+* *webchat-scroll-to-bottom-button*  
+This is the button that appears in the chat area when scrolled up. It can be clicked to scroll to the bottom.
+```CSS
+[data-cognigy-webchat-root] [data-cognigy-webchat].webchat .webchat-scroll-to-bottom-button {
+
+    background-color: rgb(88, 88, 88);
 }
 ```
 * *webchat-message-row*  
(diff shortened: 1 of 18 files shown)


sushmi21:
The chat does not auto scroll to bottom anymore. Was the scroll button implemented as a work-around for this?

sushmi21:
Scrolling works as expected with the new setting :)

sushmi21 (approved):

mastasky:
When using collated streamed messages and leaving the window, you get a sound for every incoming "chunk". We should probably only emit one chunk per "message ID".

mastasky:
In addition, when closing the webchat, we get previews for every message and markdown isnt rendered in the preview.

mastasky:
A marker on the webchat shows the number of missed messages. this marker is never reset. It remains on whatever number was reached and only ever increases.

dshire:
> When using collated streamed messages and leaving the window, you get a sound for every incoming "chunk". We should probably only emit one chunk per "message ID".

fixed

dshire:
> In addition, when closing the webchat, we get previews for every message and markdown isnt rendered in the preview.
> 
> image

Fixed:
- only the first chunk is previewed
- markdown characters are now removed. adding markdown rendering in the teaser message is blocked by updating the React version in the webchat to v18, which I won't do in this PR. created a follow up to add markdown rendering in teaser. (https://cognigy.visualstudio.com/Team%20X/_workitems/edit/91479)

mastasky (approved):
Approved, all works

dshire:
Will merge this as soon as we agree on merging the related cognigy repositories

mastasky (approved):
approved

PR #74, opened 2025-03-04
Removed relative positioning from scrollable chat area

dshire:
There is a conflict in positioning relative to the parent container between the opened date picker and the "scroll-to-bottom"-button. This PR removes the relative positioning from the scrollable area that was introduced to position the scroll-button.

see here for how it should look (and looks with this fix):

Screenshot from 2025-03-04 13-54-02
Screenshot from 2025-03-04 13-53-45

This "fix" now leads to another issue: If the input area grows (due to long text in it) AND the scroll-button is visible, it will be displayed on top of the input area. see screenshot below.

Screenshot from 2025-03-04 13-53-54

# Success criteria

- date picker is positioned correctly when open
- scrollable chat area working as expected (no double scroll bars, overflowing elements etc.)

# How to test

1. deploy webchat with date picker
2. make sure there are enough message that chat area is scrolling
3. test for breaking display

# Security

- [ ] XSS

# Additional considerations

### DIFF
--- a/src/webchat-ui/components/history/ChatScroller.tsx
+++ b/src/webchat-ui/components/history/ChatScroller.tsx
@@ -22,6 +22,7 @@ const Scroller = styled(ScrollToBottom)({
 	height: "100% !important",
 	width: "100%",
 	overflowY: "auto" as const,
+	position: "initial",
 
 	"& .hiddenAutoScrollButton": {
 		display: "none",
@@ -38,7 +39,7 @@ const ChatLog = styled.div(({ theme }) => ({
 const ScrollButton = styled("button")(({ theme }) => ({
 	position: "absolute",
 	zIndex: 10,
-	bottom: "20px",
+	bottom: "110px",
 	left: "50%",
 	transform: "translateX(-50%)",
 	backgroundColor: theme.primaryWeakColor,


sushmi21:
The scroll button is also visible inside the persistent menu. Can we hide the scroll button when the persistent menu is open?

dshire:
> The scroll button is also visible inside the persistent menu. Can we hide the scroll button when the persistent menu is open?

how do I open the persistent menu?

vj-venkatesan:
> > The scroll button is also visible inside the persistent menu. Can we hide the scroll button when the persistent menu is open?
> 
> how do I open the persistent menu?

you need to set it up in the endpoint editor .

dshire:
Okay I see it. hiding the button is not the solution here, we need to position it relatively to the scrollable chat area again. Either by changing the date picker to position itself differently, e.g. using portal, or by constantly measuring the input area and positioning the scroll button with a calculation of this measurement. But we can't do either in a quick fix here

sushmi21:
Okay, then we can merge this and see if we can find a better solution in a follow-up.

sushmi21:
Rest looks fine

sushmi21 (approved):

dshire:
> Okay, then we can merge this and see if we can find a better solution in a follow-up.

how about I quickly add a setting to disable this button? then it can be at least disabled from the endpoint in case there is issues

dshire:
or even better, we hide it by default for now

vj-venkatesan:
> or even better, we hide it by default for now

You mean the scroll button ? 
I think we can agree that this button overlapping is not as bad as the original issue. So internally in the team we agreed that is good , anything on top of this shall be done in a follow up in my opinion .

dshire:
> > or even better, we hide it by default for now
> 
> You mean the scroll button ? I think we can agree that this button overlapping is not as bad as the original issue. So internally in the team we agreed that is good , anything on top of this shall be done in a follow up in my opinion .

ok then feel free to merge and proceed. otherwise please let me know

PR #78, opened 2025-03-13
Feature/94279 scroll button setting

kwinto:
# Success criteria

- <s>Scroll Button remains unnder the input when it is multiline</s>Scroll Button is placed relative to the input disregard of its size.
- Scroll Button can be disabled by the setting (see https://cognigy.visualstudio.com/Cognigy.AI/_git/7daeebc2-e333-46d0-a819-88a7eae85978/pullrequest/40860)
- <s>The Arrow symbol on the scroll-to-bottom button is not replaced by some UTF characters when Webchat is built (this might not been fully fixes as I suspect some percularities. **Please double check after the Webchat built and released.**</s>
- 
# How to test

1. Test with settings.behavior.enableScrollButton = false;

# Security

- [ ] XSS

# Additional considerations


kwinto:
@vj-venkatesan Please check the Format job.I do not get any errors when run `npm run prettier:check`. But in the pipeline it fails...

Could it be it is not using the config from the repo? Or prettier versions mismatching?..

sushmi21:
Now, the Scroll button is not in view when it receives keyboard focus

sushmi21:
When the text in the input line is for 3 lines, the scroll button is only partly visible

sushmi21:
Hiding the "scroll to bottom" button behind the input field might negatively impact the user experience when scrollingBehavior: "scrollUntilLastInputAtTop", as this setting does not automatically scroll the user to the bottom. I think we should think of a different solution instead of just putting it behind the input field. Or at least document this behavior

src/webchat-ui/components/history/ChatScroller.tsx
```diff
@@ -154,13 +158,14 @@ const ScrollerContent = ({ children, scrolledToLastInput, setShouldScrollToLastI
...
-			{!sticky && (
+			{!sticky && !scrollButtonDisabled && (
 				<ScrollButton
 					className="webchat-scroll-to-bottom-button"
 					onClick={scrollToBottom}
 					aria-label="Scroll to bottom"
 				>
-					↓
+					{/* // Keep arrow symbol in parentheses!*/}
+					{"↓"}
```
sushmi21, line 168: This does not solve the issue. I think we need to replace the down arrow with an SVG icon or inject the HTML entity (&#8595;) using innerHTML. Anyway, it is out-of-scope for your ticket

kwinto: Yeap. Lets fix this in separate item then,

sushmi21 (approved):
Works as expected

PR #80, opened 2025-03-18
Feature/#87361 delete all conversations in history

vj-venkatesan:
# Success criteria

- Provide an option for users to delete all previous conversations directly in the webchat previous conversations.
- Ensure a menu is  provided to perform this operation
- Ensure the menu icon is only visible when previous conversations exist for the session
- New Previous Conversation options screen is visible when clicking the menu
- Delete button is visible in this screen along with text
- Ensure user confirmation alerts are displayed for the delete action.
- Provide an option for users to delete their current conversation directly in the webchat.
- Ensure user confirmation alerts are displayed for the action.
- Comply with accessibility guidelines when introducing new actions (aria-labels, etc.).

# How to test
## Delete all Conversations
1. Go to a webchat where previous conversations is enabled
2. If previous conversations exists , the menu icon should appear
3. Click that menu
4. Verify the presence of the "Delete all conversations" button in the next screen with title Coversations Options.
5. Click the "Delete all conversations" button and confirm the alert to ensure all previous conversations are deleted.
6. Ensure the user lands on the previous conversation list after deleting all conversations.

## Delete conversation
1. Go to any chat options and verify the presence of the "Delete current conversation" button in the chat options.
2. Click the "Delete current conversation" button and confirm the alert to ensure the current conversation is deleted.
3. Ensure the user lands on the empty chat (conversation history) after deleting the current conversation.

# Security

- [ ] XSS

# Additional considerations

Ensure documentation includes details on the new "Delete all conversations" button, user confirmation alerts, and compliance with accessibility guidelines.


docs/embedding.md
```diff
@@ -402,12 +402,18 @@ _These settings are NOT configurable via the Endpoint Editor in Cognigy.AI_
...
-| No network    | string | "No network connection" |
-| Reconnecting  | string | "Reconnecting..."       |
-| Reconnect     | string | "Reconnect"             |
+| Name                                  | Type   | Default                                                                                     |
+| ------------------------------------- | ------ | ------------------------------------------------------------------------------------------- |
+| Network error                         | string | "Connection Lost"                                                                           |
+| No network                            | string | "No network connection"                                                                     |
+| Reconnecting                          | string | "Reconnecting..."                                                                           |
+| Reconnect                             | string | "Reconnect"                                                                                 |
+| delete_all_conversations              | string | "delete_all_conversations"                                                                  |
```
vj-venkatesan, line 411:
```suggestion
| delete_all_conversations              | string | "Delete all conversations"                                                                  |
```

src/webchat-ui/components/WebchatUI.tsx
```diff
@@ -49,7 +49,10 @@ import {
 } from "../../webchat/helper/connectivity";
 import { HomeScreen } from "./presentational/HomeScreen";
 import { PrevConversationsList } from "./presentational/previous-conversations/ConversationsList";
-import { PrevConversationsState } from "../../webchat/store/previous-conversations/previous-conversations-reducer";
+import {
+	PrevConversationsState,
+	prevConversations,
```
vj-venkatesan, line 54:
```suggestion
	
```

src/webchat-ui/components/presentational/chat-options/DeleteConversation.tsx
```diff
@@ -0,0 +1,145 @@
...
+			});
+		}
+		setIsModalOpen(false);
+		dispatch(setShowChatOptionsScreen(false));
+	};
+
+	return (
+		<>
+			<Container>
+				<Typography variant="title1-semibold" className="webchat-rating-widget-title">
```
sushmi21, line 105: Please introduce a new class and add it to the customization doc. Our customers will use these classes to style the dialog as they need.

src/webchat-ui/components/presentational/chat-options/DeleteConversation.tsx
```diff
@@ -0,0 +1,145 @@
...
+					{config.settings.customTranslations?.delete_conversation ??
+						"Delete conversation"}
+				</Typography>
+				<DeleteButton
+					onClick={handleDeleteAllConversations}
+					aria-label="Delete all conversations"
```
sushmi21, line 111: Aria-label is not necessary for a button  which already has a text label

src/webchat-ui/components/presentational/chat-options/DeleteConversation.tsx
vj-venkatesan: ah thats right will do

sushmi21:
The font size/style used is not consistent with the ones used in rest of the places. Also, looks different from the mockup: 

Screenshot from 2025-03-19 18-10-16

Figma: 
Screenshot from 2025-03-19 18-13-24

Please use the Typography component from the chat-components repo, instead of creating further custom styles

src/webchat-ui/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+const CloseButton = styled(IconButton)(({ theme }) => ({
+	color: theme.black10,
+	borderRadius: 4,
+	marginLeft: "auto",
+
+	"&:focus": {
+		outline: `2px solid ${theme.primaryColor}`,
+		outlineOffset: 2,
+	},
```
sushmi21, line 57: It is great that we have focus outline for the close button. But for consistency, please remove the focus outline. focus-visible outline is enough for now

src/webchat-ui/components/presentational/previous-conversations/PreviousConversationsOptions.tsx
```diff
@@ -0,0 +1,154 @@
...
+	borderRadius: 12,
+	"&: focus-visible": {
+		outline: `2px solid ${theme.primaryColor}`,
+		outlineOffset: 2,
+	},
+	"&:hover:not(:disabled)": {
+		backgroundColor: theme.red10,
+		color: theme.black10,
+	},
+}));
```
sushmi21, line 73: Is it possible to reuse the primary/secondary buttons that we already use in other places? This way we can easily maintain all these buttons and we do not have to style them individually.

sushmi21:
We should make the color of the Delete/ Cancel buttons customizable. Better to add this to the endpoint settings. Something to discuss with Merlin.

sushmi21:
Focus Outline of the button is not consistent. Focus outline should have an offset and the color of the button, similar to 'Start conversation' button.
Screenshot from 2025-03-19 18-31-46

src/webchat-ui/components/Modal/Modal.tsx
vj-venkatesan: Yeah actually when the modal is open the autoFocus is set to this close button which is why I added this. 
Shouldn't a action element have focus when the Modal is open ?

src/webchat-ui/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,145 @@
...
+		return () => {
+			document.removeEventListener("keydown", handleKeyDown);
+		};
+	}, [isOpen, onClose]);
+
+	return (
+		<>
+			{isOpen && <Overlay onClick={onClose} />}
+			<CSSTransition in={isOpen} timeout={150} mountOnEnter unmountOnExit classNames="fade">
+				<StyledDialog open={isOpen}>
```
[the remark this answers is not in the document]
vj-venkatesan: I will name it webchat-dialog-root ? since its a generic modal and not just meant for delete functionality

sushmi21:
Keyboard focus is lost when I close the confirmation dialog. The focus should ideally go back to the element that triggered it, whenever possible.

sushmi21:
The keyboard focus is not trapped inside the modal.

src/webchat-ui/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,146 @@
...
+		return () => {
+			document.removeEventListener("keydown", handleKeyDown);
+		};
+	}, [isOpen, onClose]);
+
+	return (
+		<>
+			{isOpen && <Overlay onClick={onClose} />}
+			<CSSTransition in={isOpen} timeout={150} mountOnEnter unmountOnExit classNames="fade">
+				<StyledDialog className="webchat-dialog-root" open={isOpen}>
```
sushmi21, line 122: Using dialog tag is not enough for a screen-reader to differentiate between a dialog and a modal. Can you please try to use the showModal() and close() to properly toggle the state of the modal? I thing doing this will make the focus-trap work automatically (I need to check, not 100% sure)

vj-venkatesan: Yeah I actually tried it , but it applies the Modal for the entire screen and not just inside the widget. 
Might have to try adjusting the styles

sushmi21: If we can get that working, we do not have to worry about 'Esc' key handling, focus trap and focus restoration probably.

src/webchat-ui/components/presentational/chat-options/DeleteConversation.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const DeleteAnywaysButton = styled(Button)(({ theme }) => ({
+	color: theme.white,
+	background: theme.red,
+	marginLeft: "auto",
+	width: "30%",
+	"&:hover:not(:disabled)": {
+		background: theme.red10,
+		color: theme.black10,
+	},
```
sushmi21, line 44: Hover styles of the buttons does not look consistent with the rest of the buttons. For example, start conversation button

sushmi21:
The width of the cancel button can be reduced to fit the button label 
Screenshot from 2025-03-20 09-24-46

src/webchat-ui/components/Modal/Modal.tsx
sushmi21, line 122: A suggestion to improve screen reader accessibility would be to associate the title and confirmation text to dialog using aria-labelledby and aria-describedby attributes. So, while the title and body will likely be read anyway, using these 2 attributes makes sure they are always associated with the modal in the correct way.

src/webchat-ui/components/presentational/chat-options/DeleteConversation.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+const Container = styled.div`
+	display: flex;
+	flex-direction: column;
+	align-items: flex-start;
+`;
+
+const DeleteButton = styled(Button)(({ theme }) => ({
+	background: theme.red,
+	color: theme.white,
```
sushmi21, line 23: You can use the background-color and toggle between textDark & textLight, instead of setting the text to white. This will be helpful when the customer chooses 'white' or other light colors once we make the button background color customizable. (We have such an implementation for primary & secondary colors)

sushmi21:
Overlay is applied to the whole page. Instead, it should only for the webchat 
Screenshot from 2025-03-20 11-58-55

sushmi21:
The behavior works without any issues :)

src/webchat-ui/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,146 @@
...
+	}, [isOpen, onClose]);
+
+	return (
+		<>
+			{isOpen && <Overlay onClick={onClose} />}
+			<CSSTransition in={isOpen} timeout={150} mountOnEnter unmountOnExit classNames="fade">
+				<StyledDialog className="webchat-modal-root" open={isOpen}>
+					<ModalHeader>
+						<Typography
+							component="h3"
```
sushmi21, line 125: this variant already uses h3. So, we do not have to pass the component prop

src/webchat-ui/components/Modal/Modal.tsx
vj-venkatesan: I have added those attributes. Could you verify if I got them right ?

docs/embedding.md
```diff
@@ -263,16 +264,17 @@ All business hours settings require the setting `embeddingConfiguration.awaitEnd
...
+| activateTTSToggle         | boolean | `true`                                         | Activate Text-to-Speech Toggle by default. It also requires `behavior.enableTTS` setting to be `true` to make TTS active.                                                     |
+| labelTTSToggle            | string  | `"Enable text-to-speech"`                      | Label for the Text-to-Speech toggle                                                                                                                                           |
+| rating                    | object  | [Rating Section](#chat-options-rating-section) | Configure the Rating Section                                                                                                                                                  |
+| enableDeleteConversations | boolean | `false`                                        | If enabled, a delete button is provided in chat options screen. This button can be used to delete current conversations. Confirmation is required before the delete operation |
```
sushmi21, line 276: typo in 'enableDeleteConversation'

vj-venkatesan:
> There is some margin for one modal but not the other. IMO, it looks good with the margin. Also there is a full stop missing in one modal :D
> 
> Screenshot from 2025-03-26 09-05-14 Screenshot from 2025-03-26 09-04-58

hmm, I don't see this margin in my screen. 
Screenshot from 2025-03-26 09-12-10

docs/css-customization.md
```diff
@@ -1284,3 +1301,181 @@ The frame that adds the "card styles" such as background-color or box-shadow.
...
+```CSS
+[data-cognigy-webchat-root] .webchat-delete-conversation-title {
+    font-size: 16px;
+    font-weight: bold;
+    color: #333;
+}
+```
+
+- _webchat-delete-conversation-button_  
+  The general delete button style.
```
sushmi21, line 1422: Is this for the Delete button inside the modal or outside the modal? It will be nice to add it in the description here

vj-venkatesan:
@sushmi21 Delete conversation is broken after merging with David's recent changes . Fixing it

docs/css-customization.md
```diff
@@ -1284,3 +1301,181 @@ The frame that adds the "card styles" such as background-color or box-shadow.
...
+```CSS
+[data-cognigy-webchat-root] .webchat-modal-footer {
+    padding: 10px;
+    background-color: #f5f5f5;
+    border-top: 1px solid #ddd;
+}
+```
+
+- _webchat-delete-conversation-container_  
+  The container for delete conversation confirmation dialog.
```
sushmi21, line 1401: Is this the container for the dialog? What about 'webchat-modal-root' then? If this is for the section in chat options, please specify that here

vj-venkatesan: Hmm looks like it is not used . I will remove this

docs/css-customization.md
```diff
@@ -1284,3 +1300,171 @@ The frame that adds the "card styles" such as background-color or box-shadow.
...
+```CSS
+[data-cognigy-webchat-root] .webchat-modal-footer {
+    padding: 10px;
+    background-color: #f5f5f5;
+    border-top: 1px solid #ddd;
+}
+```
+
+- _webchat-delete-conversation-title_  
+  The title in delete conversation dialog.
```
sushmi21, line 1400: Can you please check where the class is used and update the decription, if need

docs/css-customization.md
```diff
@@ -1284,3 +1300,171 @@ The frame that adds the "card styles" such as background-color or box-shadow.
...
+
+```CSS
+[data-cognigy-webchat-root] .webchat-delete-all-conversation-text {
+    color: #dc3545;
+    font-weight: bold;
+}
+```
+
+- _webchat-header-delete-all-conversations-button_  
+  The button to delete all conversations in the header.
```
sushmi21, line 1462: Can please rephrase so that it is clear that you are talking about the previous conversations header

sushmi21:
When I click the 'delete all conversations' button, I do not see the list of previous conversations under the modal.

vj-venkatesan:
> When I click the 'delete all conversations' button, I do not see the list of previous conversations under the modal.

sushmi21:
I see the delete button even when there are no previous conversations
Screenshot from 2025-03-26 10-11-23

vj-venkatesan:
> The Delete icon color is not changed with 'deleteAllConversationIconColor'. I think we need to also apply the custom color to 'stroke' prop

Actually the path was using a fill I have set the path color now

kwinto:
I have a single conversation. I delete it. Then I go to the Home screen and click "Start conversation". The conversation continues (the history is not erased). I would expect it to be empty. 

You probably want to make sure the session is new in this case.

sushmi21:
The width of the Delete and Cancel buttons fit the content, like in figma.
Screenshot from 2025-03-26 10-44-25

vj-venkatesan:
> The button in the modal is not auto focused during keyboard navigation when re-opened. It does focus the first time though.

Just to confirm this is only when toggling delete all conversations dialog right ?

vj-venkatesan:
> I have a single conversation. I delete it. Then I go to the Home screen and click "Start conversation". The conversation continues (the history is not erased). I would expect it to be empty.
> 
> You probably want to make sure the session is new in this case.

@kwinto I have fixed this one . Can you please check again . 

1. Upon deleting a single conversation , a new session is created and all the existing messages are cleared .
2. Upon deleting all conversations in previous conversations screen, the current session is disconnected so if the current conversation if still sending messages won't appear again. A new session is created and when you go back to home screen and try to star conversation again

src/webchat-ui/components/presentational/previous-conversations/DeleteAllConversations.tsx
```diff
@@ -0,0 +1,72 @@
...
+		dispatch(disconnect()); // Disconnect the socket to prevent from receiving any new messages. Socket will be reconnected upon starting a new conversation
+		const storage = getStorage({
+			disableLocalStorage:
+				props.config.settings.embeddingConfiguration?.disableLocalStorage ?? false,
+			useSessionStorage:
+				props.config.settings.embeddingConfiguration?.useSessionStorage ?? false,
+		});
+		if (storage) {
+			Object.keys(storage).forEach(key => {
+				if (key.includes(userId)) {
```
kwinto, line 40: I think this one is risky and needs to be narrowed additionally down to the "webchat-client".

Webchat can run alongside other scripts on the page. You do not want to potentially wipe theirs data by the common userID in the key.

src/webchat-ui/components/presentational/previous-conversations/helpers.ts
```diff
@@ -88,13 +88,14 @@ export const getAvatars = (messages: IMessage[]) => {
 };
 
 export const sortConversationsByFreshness = (conversations: PrevConversationsState) => {
-	const sortedConversations: PrevConversationsState = Object.entries(conversations)
+	const sortedConversations = Object.entries(conversations)
+		.filter(([, session]) => !!session)
```
kwinto, line 92: the filter here is not necessary with you lates commit, right?

src/webchat/store/previous-conversations/previous-conversations-reducer.ts
```diff
@@ -4,45 +4,59 @@ import { RatingState } from "../rating/rating-reducer";
...
+const SET_CONVERSATIONS = "SET_CONVERSATIONS" as const;
 export const setConversations = (conversations: PrevConversationsState) => ({
-	type: SET_CONVERSATIONS as "SET_CONVERSATIONS",
+	type: SET_CONVERSATIONS,
 	conversations,
 });
 export type SetConversationsAction = ReturnType<typeof setConversations>;
 
-const UPSERT_PREV_CONVERSATION = "UPSERT_PREV_CONVERSATION";
+const UPSERT_PREV_CONVERSATION = "UPSERT_PREV_CONVERSATION" as const;
```
kwinto, line 24: image

kwinto (changes requested):

src/webchat-ui/components/presentational/previous-conversations/helpers.ts
vj-venkatesan: actually you are right , since it cannot be undefined I don't need to worry about it being not there . Yes I will change the interface to its original state so these changes may not be needed

src/webchat-ui/components/presentational/previous-conversations/DeleteAllConversations.tsx
vj-venkatesan: Great catch . I use the channel and user for deleting all conversations. For a single one I m using `getOptionsKey` function to narrow it down to client.session and user and url token

sushmi21 (approved):
Looks good

PR #85, opened 2025-03-27
Fix persistent history when used with teaser message

kwinto:
# Success criteria

- When teaser message arrives, it still should restore the chat history.

# How to test

1. Provide static session id. 
2. Init with:
```
teaserMessage: {
    teaserMessageDelay: 5000,
    text: "Hi there!"
},
```
3. Do not wait for a teaser message to arrive and open the chat.
4. Talk to a bot.
5. Reload.
6. Again, do not wait to a teaser message, open the chat. 
7. History should be restored.
8. Reload the page.
9. Wait for the teaser message. Click it.
10. History should be restored (previously it was not the case).


vj-venkatesan (commented):
When there are multiple conversations (Through previous conversations) and I start a new conversation and follow your steps, the teaser always takes me to the oldest conversation 

Say the following chat history
1. A (Recent)
2. B 
3. C 
Clicking teaser always goes to C , is that we want ?

kwinto:
> When there are multiple conversations (Through previous conversations) and I start a new conversation and follow your steps, the teaser always takes me to the oldest conversation
> 
> Say the following chat history
> 
> 1. C (Recent)
> 2. B
> 3. A
>    Clicking teaser always goes to A , is that we want ?

If you have a distinct session id configured, it will always open it.

vj-venkatesan:
> > When there are multiple conversations (Through previous conversations) and I start a new conversation and follow your steps, the teaser always takes me to the oldest conversation
> > Say the following chat history
> > 
> > 1. C (Recent)
> > 2. B
> > 3. A
> >    Clicking teaser always goes to A , is that we want ?
> 
> If you have a distinct session id configured, it will always open that conversation.

Ah yes that makes sense , what If i don't have a distinct session ?

kwinto:
> Ah yes that makes sense , what If i don't have a distinct session ?

It will then generate a random id for the session. So, the "Start conversation" button and the teaser message bubble would always lead to a new conversation.

sushmi21 (approved):

vj-venkatesan (approved):

PR #154, opened 2025-08-07
Create draft publish.yml file

saeb-cognigy:
# Success criteria

- A new pipeline called "Published" should be created
- The pipeline should publish the NPM package automatically after releasing a new version with the "latest" tag

# How to test
Same steps from the "Success criteria"

# Security

- [ ] XSS

# Additional considerations

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,18 +17,18 @@ jobs:
             - name: Get latest release version
               id: get-version
               run: |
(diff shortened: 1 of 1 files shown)


kwinto:
We might be digging in the wrong direction - it looks like there is no way to trigger an action on release EDIT...

https://chatgpt.com/share/68950ade-be14-8000-a8fe-e43b00346b3c

We can go with manual trigger, this would also be a good solution.

PR #197, opened 2025-10-30
Fix overscroll behavior and scroll leak for webchat

sushmi21:
# Success criteria

- When overscrolling a scrollable area, like chat-log, previous-conversations, etc the background body should not scroll
- When in mobile device, scrolling inside webchat should not scroll the body with or without the mobile keypad open
- Should be able to zoom in and out with pinch gestures with and without the webchat open

# How to test

1. Add overflowing content to the body of the index.html
2. Chat with the bot to create overflowing chat log
3. Test if over-scrolling inside log, scrolls the body
4. Test the webchat using a mobile phone and see if the body does not scroll when scrolling inside the webchat (like scrolling inside input fields with mobile keypad open, chat log, previous conversation list, etc)
5. Also, test the pinch zoom in and out gestures

# Security

- [ ] XSS

# Additional considerations

### DIFF
--- a/src/webchat-ui/components/WebchatUI.tsx
+++ b/src/webchat-ui/components/WebchatUI.tsx
@@ -696,12 +696,12 @@ export class WebchatUI extends React.PureComponent<
 		window.removeEventListener("resize", this.handleResize);
 	}
 
-	handleResize = () => {
+	handleResize() {
 		const isMobile = isMobileViewport();
 		if (isMobile !== this.state.isMobile) {
 			this.setState({ isMobile });
 		}
-	};
+	}
 
 	private iconAnimationIntervalHandle: ReturnType<typeof setInterval> | null = null;
 


vj-venkatesan:
Best case scenario is working good. When I scroll the chat log , the underlying body doesn't scroll. However the auto scroll behaviour we have when I open a conversation seems to scroll the background . I tested that scenario in my phone . I hope the screen recording helps. 

Its not big UX issue , however if there is way to fix it would be nice to do a follow up

https://github.com/user-attachments/assets/e3aaa3ec-9cdc-4654-83a1-1128b6360bc8

sushmi21:
> Best case scenario is working good. When I scroll the chat log , the underlying body doesn't scroll. However the auto scroll behaviour we have when I open a conversation seems to scroll the background . I tested that scenario in my phone . I hope the screen recording helps.
> 
> Its not big UX issue , however if there is way to fix it would be nice to do a follow up
> 
>  ScreenRecording_12-04-2025.10-56-57_1.MP4

I can see the issue. I think the scrolling of the page happens when the keypad opens on input focus. I will create a follow-up for this.

vj-venkatesan (approved):

PR #300, opened 2026-08-06
fix(a11y): announce the AI-agent notice once per new conversation (CGY-3519)

sushmi21:
## What

Announces the AI-agent notice ("You're now chatting with an AI Agent." / configured `AIAgentNoticeText`) to screen-reader users when the chat screen appears — once per **brand-new** conversation, always **before** any message announcement.

The static notice text at the top of the chat log can't announce itself: it renders together with the screen (no live-region change), focus lands in the message input *past* it in reading order, and `aria-describedby` on the input would be re-read on every re-focus.

## How

- **Own live region, coordinated with the message region** (`ScreenReaderLiveRegion`): the notice renders as an "intro" into a dedicated sibling region (`#webchatAIAgentNoticeLiveRegion`). A shared region would let the next message announcement replace the notice's node — NVDA silently drops a queued polite announcement whose node leaves the DOM. While the intro is pending, message announcements hold, so mutations always reach the screen reader in intro-first order. The intro is deferred 600&nbsp;ms past the screen's focus utterances.
- **Once per brand-new session** (`computeNoticeSession`, `src/webchat-ui/utils/ai-agent-notice.ts`): new-ness is evaluated against a pre-switch snapshot of `prevConversations` (a new session gets upserted into that map by auto-inject in the same batch). Reopening a previous conversation or navigating away and back stays silent; the page load's first conversation keeps its announce key across the connect that assigns its session id. Announced keys live on the `WebchatUI` instance because the live region remounts on every screen navigation.
- **Held while the disconnect overlay is open** (session switches reconnect the socket): the overlay's dialog/focus utterances would cancel the notice; it becomes pending again on close.
- `HomeScreenAnnouncer` generalized to `ScreenAnnouncer`; screen-visibility flags unified in `getScreenVisibility` (`src/webchat-ui/utils/screen-visibility.ts`), the single source for the announcers and `renderRegularLayout`.

## Tests & docs

- New `AI-agent notice announcement (CGY-3519)` describe block in `cypress/e2e/screenReaderLiveRegion.cy.ts`: default and configured text, ordering vs. a simultaneous message, no re-announce on return, re-announce for a new conversation from previous conversations, silence when reopening a previous conversation, and the disabled setting.
- Pattern documented in `docs/accessibility.md`.

## How to test

1. `npm run dev`, open http://localhost:8787 against an endpoint, and start a screen reader (NVDA/VoiceOver).
2. Open the webchat and start a conversation → after the focus utterance you should hear **"You're now chatting with an AI Agent."**, and only then any incoming message.
3. Go back to the home screen and return to the chat → no repeat announcement.
4. Send a message, then open **Previous conversations** and start a **new** conversation → the notice is announced again (after the reconnect overlay closes).
5. Reopen the **same** previous conversation instead → stays silent.
6. Config variations: set `settings.behavior.AIAgentNoticeText` → the custom text is announced; set `settings.behavior.enableAIAgentNotice: false` → nothing is announced.
7. QA - Please test for any regressions

Without a screen reader, you can watch the announcements land in the off-screen live region: inspect `#webchatAIAgentNoticeLiveRegion` (notice) and `#webchatMessageContainerScreenReaderLiveRegion` (messages) in devtools.


kwinto (commented):
Scoped to this PR's own 4 commits (`b077ac2..a049f336`) — the rest of the diff is #299, still open.

**Verified locally at `a049f336`:** `npm run lint:a11y` clean (1 pre-existing warning), `prettier --check` clean on all touched files, `eslint` clean on the new files, `tsc --noEmit` produces no new errors (the repo has many pre-existing ones).

## Verdict

The approach is right, and unusually well-reasoned for this problem class. Three things I checked specifically and found correct:

1. **The pre-switch `prevConversations` snapshot is genuinely necessary.** In `previous-conversations-middleware.ts:47-58`, `setOptions` (which changes `currentSession`) is followed synchronously by `triggerAutoInject` → `SEND_MESSAGE` → `upsertPrevConversation`. Under `createRoot` (`src/webchat-embed/index.tsx:82`) React 18 batches all of it into one commit, so `componentDidUpdate` can see the new session *already present* in `this.props.prevConversations`. `prevProps.prevConversations` is the only correct source. This is easy to "simplify" wrongly later — the comment at `WebchatUI.tsx:543` earns its place.
2. **The `isInforming` ⇄ `showInformationMessage` substitution in `getScreenVisibility` is safe.** The docstring claim ("truthy iff") holds: `isInformingOutOfBusinessHours` → `isOutOfBusinessHours` → checks `businessHours.enabled` (`src/webchat/helper/businessHours.ts:3`), so every `isInforming` branch also satisfies one of the three `informMessage` branches, all of which have non-empty fallback text.
3. **Dropping the `open` check from the home-screen announcer condition is safe** — `<ScreenAnnouncer>` is inside `{open && …}` at `WebchatUI.tsx:1321`.

Two behavioral issues below are worth addressing before merge; the rest are nits.

## 1. The notice can be permanently swallowed if the user leaves the chat screen inside the 600 ms window

During back-to-home, `getRegularLayoutContent()` keeps returning the chat branch (it's the fallthrough), `FreezeOnExit` holds the same element tree, and `unmountOnExit` only fires after the 500 ms exit — so `ScreenReaderLiveRegion` stays **mounted** inside a wrapper that now carries `aria-hidden="true"` + `inert` (`WebchatUI.tsx:1820-1829`).

If the intro timer fires in that window, it commits text into an `aria-hidden`/`inert` region (silence) **and still calls `onIntroAnnounced`**, which adds the key to `announcedNoticeKeys`. Since that set outlives the remount by design, the notice is never announced again for that session.

Window: navigate away between ~100 ms and ~600 ms after the chat screen appears (before 100 ms the unmount clears the timer first; after 600 ms it already announced). Narrow, but reachable — and it's the same failure mode #299 fixed for messages.

Fix: at fire time, only commit and mark-announced if the region is still exposed. With a ref on the region, `!el.closest('[inert], [aria-hidden="true"]')` is enough; if it's hidden, skip both `setIntroMessage` and `onIntroAnnounced` so it re-announces on the next real visit. Gating on `showChatScreen` in the parent won't work — `FreezeOnExit` freezes props, so `introText` stays defined through the exit.

## 2. Page reload of a persisted conversation re-announces

At mount `options.sessionId` is `""` — `setInitialSessionId` writes to `config.initialSessionId`, not `options` (`src/webchat/components/Webchat.tsx:82`) — so `computeNoticeSession` yields `{id:"", isNew:true, announceKey:""}`. When the connect dispatches `setOptions`, `optionsMiddleware` dispatches `setConversations(...)` and `resetState(persisted)` **before** `next(action)` applies `SET_OPTIONS` (`src/webchat/store/options/options-middleware.ts:32-57`); React batches all three, so `prevProps.prevConversations` is still `{}`. `isNew` stays `true`, `isFirstConnectOfNewConversation` keeps `announceKey: ""`, and the notice announces once the chat screen appears — on a conversation that is visibly continuing with restored history.

Note this is *not* a `getAllConversations` exclusion: `currentSessionId` is accepted but never used in that function (`src/webchat/helper/storage.ts:24-56`), so the restored session *is* in the map — it's the batching that hides it.

Either fix it (a non-empty restored `messageHistory` is a clean "not brand-new" signal) or state in `docs/accessibility.md` that a reload counts as a fresh announcement. Right now the doc says "once per **brand-new** conversation", and `persistent-history.cy.ts` covers a flow this contradicts. My read is it should be silent — the user is mid-conversation, and the visible notice is still there to read.

## 3. `introKey ?? ""` can deadlock message announcements

`introKey` is optional, but `ScreenReaderLiveRegion.tsx:61-62` coerces it to `""` when storing. If any caller ever passes `introText` without `introKey`, `announcedIntroKey` becomes `""` while `introKey` stays `undefined`, so `introPending` is permanently `true` — and since the effect deps don't change, nothing recovers. **All message announcements stop for that mount.**

Not reachable today (`WebchatUI` always passes `announceKey`, a string), but the failure is silent and total. Make the pair inseparable — `intro?: { key: string; text: string }` — or drop the `??` and let the types force it.

## 4. Test timing margins are thin

- `"announces the notice before a message that arrives at the same time"` (`cypress/e2e/screenReaderLiveRegion.cy.ts:65`): the intro timer starts during `beforeEach`'s `startConversation()`, not in the test body. The `should("be.empty")` assertion lands at roughly command-overhead + 450 ms against a 600 ms deadline — a margin of ~100-140 ms, and `be.empty` passes on first check so retries can't save it. #299 already moved a comparable spec to `cy.clock()`; same treatment here would make it deterministic.
- The three negative assertions (`not.contain.text`) pass because navigation unmounts and clears the region, not because nothing was announced. That's still a valid signal, but it's load-bearing and undocumented — worth a comment so a future "keep the region mounted" change doesn't turn these green-forever.
- Per the repo's Definition of Done, this spec has no `cy.checkA11yCompliance("[data-cognigy-webchat-root]")` block. The new surface is `sr-only` so axe coverage elsewhere probably suffices — flagging for the checklist, not as a real gap.

## Nits

- `docs/accessibility.md`: "(the PostNL acceptance criterion)" names a specific customer in a general pattern doc. State the requirement generically.
- Same file: the new pattern paragraph is two ~6-line sentences. The reasoning is worth keeping, but it needs breaking up.
- `ScreenAnnouncer` is `HomeScreenAnnouncer` renamed with no behavior change, and its docstring now spends a paragraph explaining what it is *not* used for. Since the generalization it was extracted for got reverted in `d731bb3f`, consider whether the rename still pays for itself.

---

kwinto (commented):
_Duplicate submission — see the review above (same content)._

sushmi21:
Thanks for the thorough review — all four findings plus the nits are addressed in 0966120f.

**1. Notice swallowed during the 600 ms window** — fixed with the fire-time exposure check you suggested: `SrOnlyLiveRegion` now forwards a ref, and the intro timer bails when the region sits inside `[inert]`/`[aria-hidden="true"]`, skipping both the commit and `onIntroAnnounced` so the notice re-announces on the next real visit. (Agreed that gating on `showChatScreen` can't work under `FreezeOnExit`.)

**2. Reload of a persisted conversation** — went with silent, matching the doc's "brand-new" wording. One refinement over the suggested signal: "non-empty `messageHistory` at first connect" can't distinguish a restore from a brand-new conversation with `startBehavior: "injection"` — the auto-injected message lands in `messageHistory` in the same React commit as `SET_OPTIONS` (autoinject-middleware fires synchronously right after it). Instead, a `hasRestoredPersistedHistory` flag is set only where a restore unambiguously happens (the `RESET_STATE` handler, on a non-empty restore) and consulted only for the page load's first connect, so later restores and post-switch new conversations are unaffected. Documented in `docs/accessibility.md` and covered by a new reload-silence spec.

**3. `introKey ?? ""` deadlock** — the pair is now a single `intro?: { key: string; text: string }` prop; the coercions are gone and the invalid state is unrepresentable. The effect depends on destructured primitives so the per-render object identity doesn't restart the announce timer.

**4. Test timing** — the intro-order spec now rebuilds the webchat with `cy.clock(Date.now(), ["setTimeout", "clearTimeout"])` installed before the chat screen mounts and asserts at ticks 599/600/700. The unmount-dependent negative assertions got explanatory comments, and the spec gained the Definition-of-Done `Accessibility (WCAG 2.2 AA)` block.

**Nits** — customer name removed and the two long sentences split in `docs/accessibility.md`; `ScreenAnnouncer` reverted to `HomeScreenAnnouncer` (the generalization it anticipated was indeed reverted, so the rename no longer paid for itself).

Verified locally: `lint:a11y` clean (same single pre-existing warning), prettier clean, `tsc --noEmit` produces an identical error set to the baseline (stash-diffed). Cypress runs in CI.

kwinto (approved):

mboyech (approved):

PR #303, opened 2026-08-21
fix(a11y): home-screen exit focus fallback + review follow-ups (CGY-3278)

sushmi21:
Follow-ups from the review of #302 (kwinto's two review comments). Based on `fix/CGY-3278-homescreen-aria-hidden`.

## Changes

1. **Focus fallback on home-screen exit (SC 2.4.3).** `inert` on the hiding home screen blurs the control that triggered the exit (e.g. the just-clicked *Start conversation* button). Every exit except one re-focuses on its own (header title via `autoFocusScreenTitle`, message input autofocus, back-to-home's explicit focus); with `widgetSettings.disableInputAutofocus: true`, entering the chat screen left focus on `document.body`, so Tab restarted at the top of the host page. Pre-inert, the same path parked focus on the hidden `aria-hidden` button — a screen-reader void — so the gap predates #302; `inert` only changed the failure mode. `WebchatUI.componentDidUpdate` now checks 450 ms after the home screen hides (after BaseInput's and the Header's 200 ms timers, matching the back-to-home timing) and moves focus to the header title only if it is still on `document.body` or inside the hidden home screen.
2. **Regression specs** for both review findings: the fallback focuses the header title, and reopening the widget onto the chat screen (home screen mounts already hidden) never places focus inside `.webchat-homescreen-root` — on `main`, the home screen's mount effect focused the hidden close button whenever input autofocus was disabled.
3. **Shared `InertProps` type** (`src/webchat-ui/utils/inert-props.ts`) replaces the duplicated React-18 `inert?: string` styled-prop workaround in `HomeScreenRoot` and `RegularLayoutContentWrapper`.
4. **Split the chained `have.attr` assertions** in the first CGY-3278 spec (re-subject footgun).
5. **Doc sync**: `CLAUDE.md`, `.claude/skills/wcag-component/SKILL.md`, `.github/copilot-instructions.md`, and `docs/accessibility.md` now describe the two-list `getKeyboardFocusableElements` contract, the `inert`-based hide/show pattern, and the two SC 2.4.3 behaviors above.

## How to test

Dev server: `npm run dev` (:8787), or use `dist/index.html` with a home-screen-enabled endpoint. Set:

```js
settings: { widgetSettings: { disableInputAutofocus: true } }
```

1. **Exit-focus fallback:** open the widget → home screen shows. Tab to **Start conversation** and press Enter. The chat screen slides in; ~450 ms later focus is on the header title (check `document.activeElement` in the console — it must not be `<body>`). Press Tab: focus continues inside the widget instead of restarting at the top of the host page. Repeat via a conversation starter button.
2. **No-regression on the self-focusing exits:** same flow with `disableInputAutofocus` removed → the message input receives focus as before. Home → *Previous conversations* → header title focuses (existing behavior). Back button → home-screen close button focuses (existing behavior).
3. **Reopen onto chat screen:** with `disableInputAutofocus: true`, start a conversation, minimize via the toggle, reopen via the toggle. Focus lands on the chat header back button — not the hidden home-screen close button (which it did on `main`).
4. **CI:** the two new specs in `cypress/e2e/homeScreen.cy.ts` (`Accessibility (WCAG 2.2 AA)` block) cover 1 and 3.


kwinto (dismissed):
Checked this against my follow-ups on #302 — all four are addressed, and correctly:

- The **exit-focus fallback** is guarded the way it should be: it only acts if focus is still on `document.body` or inside the hidden root, it runs after both 200 ms timers (`BaseInput` autofocus, `Header` `autoFocusScreenTitle`) so the self-focusing exits are untouched, the previous timeout is cleared before re-arming, and it's torn down in `componentWillUnmount`. The `webchatWindowRef` null-guard also covers the "user minimizes within 450 ms" case, since `WebchatRoot` only mounts while `open`.
- `.webchat-header-title` has an unconditional `tabIndex={-1}`, so the target is reliably focusable on every screen — good.
- Both regression specs are meaningful (the first fails pre-fix), `InertProps` removes the duplication, the chained `have.attr` is split, and the four doc files now describe the two-list contract accurately.

Approving. Three small things, all optional:

**1. `active === document.body` can match a user action, not just the inert blur (low).**
If the user clicks a non-focusable area of the *host page* within the 450 ms window, `document.activeElement` becomes `<body>` and the widget will pull focus into its header title — focus-stealing from a deliberate user action. Tighter condition: record in `getSnapshotBeforeUpdate` whether `document.activeElement` was inside `.webchat-homescreen-root` *before* the commit that applies `inert` (that's the only point where it's still observable), and fire only then. Narrow window, so fine to leave as-is if you'd rather not add a snapshot hook.

**2. The fallback silently no-ops while the xApp overlay is open (very low).**
The `Header` is gated behind `!isXAppOverlayOpen`, so `.webchat-header-bar .webchat-header-title` doesn't exist and focus stays on `<body>`. A `?? getKeyboardFocusableElements(webchatWindowEl).firstFocusable` second choice would make the fallback total.

**3. Spec nit — assert the positive target.**
`cy.focused().closest(".webchat-homescreen-root").should("not.exist")` fails with a confusing "requires a DOM element" error rather than a clean assertion if focus is ever lost to `<body>` (`cy.focused()` yields `null` then). Asserting what the PR description promises — focus on the header back button — is both stronger and clearer on failure.

Merge note: this is stacked on `fix/CGY-3278-homescreen-aria-hidden`, so #302 needs to land first (GitHub will retarget this to `main` automatically).

sushmi21:
Thanks @kwinto — all three follow-ups are in (28546c49), plus the retarget:

1. **Snapshot gate + pointerdown disarm.** `getSnapshotBeforeUpdate` now records whether focus was inside `.webchat-homescreen-root` just before the update that applies `inert`, and the fallback is only armed in that case — so programmatic exits never steal focus. Since the snapshot alone doesn't cover your exact scenario (user clicks Start, *then* clicks a non-focusable host-page area inside the 450 ms window — focus was inside the home screen at snapshot time, yet lands on `<body>`), a capture-phase `document` `pointerdown` listener additionally disarms the pending fallback, so a deliberate click always wins. The listener is removed when the timer fires, on re-arm, and in `componentWillUnmount`.

2. **Total fallback.** The focus target is now `header title ?? getKeyboardFocusableElements(webchatWindowEl).firstFocusable`, so it still lands somewhere while the xApp overlay hides the Header.

3. **Positive spec assertion.** The reopen spec now asserts `cy.focused().should("have.class", "webchat-header-back-button")` — the window's first focusable on the chat screen — instead of the `.closest()` chain.

`docs/accessibility.md` describes the arm/disarm conditions. On your merge note: #302 landed squash-merged, so I retargeted this PR to `main` and rebased onto it — the diff now contains only this PR's changes.

kwinto (approved):

PR #304, opened 2026-08-21
fix(security): correct xApp cross-frame origin check (WCH-SI10-004)

smohite-nice:
## Summary

- **Finding:** WCH-SI10-004 — HIGH — Cross-frame `postMessage` origin check semantically backwards and bypassable
- **Control:** SI-10 / AC-4
- **Jira:** [CSA-97605](https://nice-ce-cxone-prod.atlassian.net/browse/CSA-97605)
- **Confluence:** [WCH-SI10-004 remediation page](https://cognigy.atlassian.net/wiki/spaces/Engineering/pages/2745073678)

### The bug

`XAppOverlay.tsx` checked cross-frame `postMessage` origins with:

`	s
if (url.startsWith(event.origin) === false) { return; }
`

This is semantically backwards — a domain that is a string-prefix of the configured URL (e.g. `https://xapp.cognigy.a` for `https://xapp.cognigy.ai/form`) passes the check despite being a different origin. A passing message can inject into the chat log or close the overlay.

### The fix

`	s
let urlOrigin: string;
try {
    urlOrigin = new URL(url).origin;
} catch {
    return; // empty or invalid url — reject
}
if (urlOrigin !== event.origin) { return; }
`

Also adds `url` to the `useEffect` dependency array to prevent a stale-closure when the xApp URL changes.

## Test plan

- [x] New: origin match → overlay closes (`closeOnSubmit: true`)
- [x] New: origin mismatch → overlay stays open
- [x] New: wrong message type ignored even with matching origin
- [x] New: `cy.checkA11yCompliance` on xApp overlay (WCAG 2.2 AA)
- [x] All 4 pre-existing xApp overlay tests unchanged

### Security
- Fixed xApp `postMessage` origin check — replaced `url.startsWith(event.origin)` with `new URL(url).origin === event.origin` to close string-prefix bypass (WCH-SI10-004, SI-10/AC-4)

[CSA-97605]: https://nice-ce-cxone-prod.atlassian.net/browse/CSA-97605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/messages/xAppsOverlay.cy.ts
+++ b/cypress/e2e/messages/xAppsOverlay.cy.ts
@@ -41,3 +41,129 @@ describe("xApps Overlay", () => {
 		});
 	});
 });
+
+/**
+ * WCH-SI10-004 — Cross-frame origin check fix.
+ *
+ * The previous check `url.startsWith(event.origin)` compared the raw URL string
+ * against the sender origin, which is semantically backwards. A domain that is a
+ * string-prefix of the configured xApp URL (e.g. "https://xapp.cognigy.a" for
+ * "https://xapp.cognigy.ai/form") could pass the check despite being a completely
+ * different origin.
+ *
+ * The fix uses `new URL(url).origin === event.origin` which compares the canonical
+ * scheme+host+port extracted from both sides — the correct cross-origin boundary.
+ *
+ * Test strategy: in the Cypress environment, postMessages sent from the test page
+ * carry origin "http://localhost:8787". Setting the xApp URL to that same origin
+ * lets us exercise the acceptance path; using a different URL origin exercises the
+ * rejection path. Both cases are observable via the closeOnSubmit behaviour.
+ */
+describe("postMessage origin validation (WCH-SI10-004)", () => {
+	beforeEach(() => {
+		cy.visitWebchat().initMockWebchat().openWebchat().startConversation();
+	});
+
+	it("closes overlay when postMessage origin exactly matches the xApp URL origin", () => {
+		// xApp URL uses localhost:8787 — the same origin as the Cypress test runner —
+		// so postMessages dispatched from the test page have a matching origin.
+		cy.receiveMessage(null, {
+			_cognigy: {
+				_app: {
+					overlaySettings: {
+						autoOpen: true,
+						closeOnSubmit: true,
+						feedbackMessage: "",
+						screenTitle: "Local xApp",
+						sendEventOnCloseIconClick: false,
+						showCloseIcon: false,
+					},
+					url: "http://localhost:8787/xapp-test",
+				},
+			},
+		});
+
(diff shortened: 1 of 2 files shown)


kwinto (approved):
Approving — this is a strict improvement over `url.startsWith(event.origin)` and I could not construct any customer-facing regression from it.

**Verified as safe:** custom domains / CNAMEs / self-hosted origins are unaffected (the comparison is against the bot-supplied `url`, not a hard-coded `cognigy.ai` allow-list); relative, protocol-relative and empty URLs already failed the old check and now fail cleanly via the `try/catch`; WHATWG canonicalisation (default ports, uppercase hosts) makes the new check slightly *more* permissive, not less. Also confirmed this is the only `addEventListener("message", …)` in `src/`, so there is no inconsistent second handler.

Nice catch on the `useEffect` dependency array — adding `url` and `feedbackMessage` fixes a real stale-closure bug where opening a second xApp would validate against the previous xApp's origin.

Three non-blocking follow-ups, none of which should hold this up:

**1. The new tests pass against the pre-fix code.** All three cases resolve identically under both implementations:

| Test | `url` | `event.origin` | old `startsWith` | new `===` |
|---|---|---|---|---|
| accept | `http://localhost:8787/xapp-test` | `http://localhost:8787` | true | true |
| reject | `https://example.com/xapp-form` | `http://localhost:8787` | false | false |

So a revert of the one-line fix keeps CI green. A discriminating case needs `event.origin` to be a strict string-prefix of `url` while being a different origin — e.g. `url: "http://localhost:8787.evil.example/xapp"`, where `startsWith("http://localhost:8787")` is `true` but the origins differ.

**2. No `event.source` check.** The listener is on the host `window` and only validates `origin`. Where the xApp is served from the same origin as the embedding page, any same-origin script or iframe can post `{ type: "x-app-submit", success: true }` and inject a `source: "user"` message via `addMessage` plus force-close the overlay. A `useRef` on the `<Iframe>` plus `if (event.source !== iframeRef.current?.contentWindow) return;` closes this.

**3. Opaque origins compare equal.** `new URL(x).origin` returns the *string* `"null"` for `data:`, `about:`, `file:` and non-special schemes, and `event.origin` is also `"null"` for sandboxed/`srcdoc` frames — so they match. `url` comes from bot output (`_cognigy._app.url`) and is never scheme-validated, and the old `startsWith` happened to reject these, so this is a new acceptance path. Worth adding `if (u.protocol !== "https:" && u.protocol !== "http:") return;` after the parse. Note this is *not* caught by #307's guard either, since `new URL("data:…")` does not throw and its origin is not equal to `window.location.origin`.

Given #307 touches the same lines, these are cheap to fold into whichever of the two rebases second.

PR #309, opened 2026-08-25
fix(security): add deny-list to customAllowedHtmlTags in sanitizeHTML (WCH-SI10-002)

smohite-nice:
Adds a hard deny-list to `sanitizeHTML()` so that dangerous HTML tags can never be re-enabled through the `customAllowedHtmlTags` endpoint configuration, regardless of what a tenant supplies.

**Scope — user-typed input and PrivacyNotice text only.**
`sanitizeHTML()` in this repo is called on two surfaces:
1. **User-typed messages** — `message-middleware.ts:123` (SEND_MESSAGE path), when `disableTextInputSanitization` is `false`
2. **PrivacyNotice body** — `PrivacyNotice.tsx:58`, sanitizing the tenant-configured notice text

**Bot and agent messages are out of scope.**
Incoming bot/agent messages flow through `RECEIVE_MESSAGE → addMessage() → @cognigy/chat-components`, which never calls `sanitizeHTML()` in this repo. Tags like `style` and `iframe` that customers use for bot message styling and embedding are handled entirely in `@cognigy/chat-components` and are unaffected by this change.

**The `<script>` tag that loads `webchat.js` is unrelated.**
Blocking `<style>` or `<script>` in `sanitizeHTML()` has zero effect on how the widget loads or renders. The webchat's own styling uses Emotion CSS-in-JS which injects `<style>` elements directly — DOMPurify never touches the widget's own markup. The deny-list only strips those tags when a user **types** them into the chat input.

**FedRAMP SI-10 / AC-3 remediation (WCH-SI10-002 — partial):**
`customAllowedHtmlTags` previously replaced `ALLOWED_TAGS` wholesale with no validation. A tenant could set `customAllowedHtmlTags: ["script"]` and make every session on that endpoint XSS-vulnerable in a single config change.

Changes:
- `src/webchat/helper/sanitize.ts` — adds `ALWAYS_BLOCKED_TAGS` (exported `Set` of 12 tags) and filters the tenant-supplied list against it before passing to DOMPurify. The filter is case-insensitive (`.toLowerCase()`) to prevent `SCRIPT`/`Script` bypasses. Also adds `FORBID_TAGS: [...ALWAYS_BLOCKED_TAGS]` to the base DOMPurify config so the block is unconditional on both the default and custom-tags code paths.

Always-blocked tags (regardless of tenant config):
`script`, `iframe`, `object`, `embed`, `applet`, `frame`, `frameset`, `meta`, `base`, `link`, `style`, `form`

Safe tags in `customAllowedHtmlTags` (e.g. `["p", "b", "span"]`) continue to work unchanged. Behaviour when `customAllowedHtmlTags` is not set is unchanged.

Jira: [CSA-97603](https://nice-ce-cxone-prod.atlassian.net/browse/CSA-97603)

# Success criteria

- `customAllowedHtmlTags: ["script"]` no longer allows `<script>` tags through `sanitizeHTML`
- All 12 always-blocked tags are filtered regardless of casing in the tenant-supplied array
- Tenants using `customAllowedHtmlTags` with safe tags (e.g. `["p", "b", "span"]`) are unaffected
- Bot message rendering is unaffected — customers using `style` or `iframe` in bot messages see no change
- `customAllowedHtmlTags` not set → existing behaviour unchanged

# How to test

1. Run `npm run build && npm test` — all Cypress tests including the new `customAllowedHtmlTags` tests pass
2. Set `customAllowedHtmlTags: ["script"]` in endpoint config, send a message containing a script tag — verify no `<script>` element appears and no script executes
3. Set `customAllowedHtmlTags: ["SCRIPT"]` (uppercase) — same result (case-insensitive block)
4. Set `customAllowedHtmlTags: ["p", "b"]` — verify those safe tags still work
5. Receive a bot message with inline styles or an iframe — verify it renders correctly and is unaffected

# Security

- [ ] XSS

# Accessibility (WCAG 2.2 AA)

- [x] No accessibility implications (non-UI change)

# Additional considerations

The `customAllowedHtmlTags` row in `docs/embedding.md` has been updated to list the 12 always-blocked tags. Bot message rendering (and its allow-list) lives in `@cognigy/chat-components` and is not changed by this PR.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

[CSA-97603]: https://nice-ce-cxone-prod.atlassian.net/browse/CSA-97603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kwinto (changes requested):
Please add a comment about this black list into embedding doc to the related row.

kwinto (changes requested):
Thanks for adding the docs row — that addresses my earlier comment. The deny-list itself is well built: case-insensitivity is handled, and using `FORBID_TAGS` *plus* a pre-filter is genuine defence-in-depth rather than a single gate (DOMPurify applies `FORBID_TAGS` over `ALLOWED_TAGS`, so the ordering is correct). Two things need fixing before merge.

---

**1. (Blocking) Malformed `customAllowedHtmlTags` now throws — blank widget.** `sanitize.ts:264-265`

```ts
customAllowedHtmlTags.filter(tag => !ALWAYS_BLOCKED_TAGS.has(tag.toLowerCase()))
```

This assumes an array of strings, but the value arrives from remote endpoint JSON / embed-time overrides and is deep-merged with no type validation in `config-reducer.ts`. I ran each shape against the exact filter logic:

| value | result |
|---|---|
| `["p", null]` | `TypeError: Cannot read properties of null (reading 'toLowerCase')` |
| `["p", 1]` / `[{}]` | `TypeError: tag.toLowerCase is not a function` |
| `"p,br"` (string) | `TypeError: customAllowedHtmlTags.filter is not a function` |

**All four are handled gracefully on `main`** — DOMPurify's `addToSet` tolerates non-strings and non-arrays, so this is a new failure mode introduced here. Impact: `PrivacyNotice.tsx:58` calls `sanitizeHTML` inside `useMemo` *during render*, and there is no `ErrorBoundary`/`componentDidCatch` anywhere in `src/`, so React unmounts the entire widget — blank webchat. On the `SEND_MESSAGE` path (`message-middleware.ts:123`) it throws out of `dispatch` and breaks message sending.

```ts
const list = Array.isArray(customAllowedHtmlTags) ? customAllowedHtmlTags : [];
const safeTags = list.filter(
    tag => typeof tag === "string" && !ALWAYS_BLOCKED_TAGS.has(tag.toLowerCase().trim()),
);
```

(The `.trim()` also closes the cosmetic `" script "` gap — harmless today since DOMPurify never matches a tag name containing spaces, but it belongs in the same fix.)

**2. (Blocking, docs) `FORBID_TAGS` changes the default path too.** `sanitize.ts:251`

`FORBID_TAGS` is added to the base `config`, so all 12 tags are stripped even when `customAllowedHtmlTags` is unset — they were previously in `allowedHtmlTags`. The success criteria say *"`customAllowedHtmlTags` not set → existing behaviour unchanged"*, which the PR's own test (`FORBID_TAGS also blocks form in the default config`) contradicts, and the `docs/embedding.md` edit only documents the block under the `customAllowedHtmlTags` row. Real blast radius is small, but the description and docs should say what the code does. Landing #310 first makes this a genuine no-op and resolves it cleanly.

---

**Non-blocking:**

- **Test assertions are vacuous for 8 of the 12 tags.** `customAllowedHtmlTags.cy.ts:113-137` only sends `script`/`iframe`/`form`/`object` markup but loops `expect(text).not.to.contain('<' + tag)` over all 12 — the assertions for `embed`, `applet`, `frame`, `frameset`, `meta`, `base`, `link`, `style` pass regardless of implementation. Please include markup for every tag, and add cases for the malformed shapes above.
- **Reconcile with #310.** #310 considers `body` and `noframes` dangerous enough to delete from the default allow-list, but they aren't in `ALWAYS_BLOCKED_TAGS` — so after both land, a tenant can re-enable exactly those two via config. Suggest adding them here.
- **A one-time `console.warn`** listing dropped tags would make support cases diagnosable; silent-drop is correct behaviour but opaque to whoever configured it.
- **Merge order.** #309 and #310 auto-merge cleanly (`git merge-tree`, exit 0) and the result is coherent. Recommend **#310 first, then this** — otherwise `main` temporarily ships an `allowedHtmlTags` constant that contradicts effective behaviour.

---

**On CSA-97603 closure.** The PR body notes chat-components is out of scope, and I agree for this diff — but the docs now say these tags are blocked *"regardless of this setting"*, and that isn't true on the surface where tenant-influenced content actually renders. `@cognigy/chat-components` reads the **same** `config.settings.widgetSettings.customAllowedHtmlTags` (via `useSanitize` → `sanitizeHTMLWithConfig`) and passes it straight to `ALLOWED_TAGS` with no deny-list, then renders through `dangerouslySetInnerHTML`. Its own defaults still allow `iframe`, `object`, `embed`, `base`, `meta`, `link`, `form`, `style` plus `srcdoc`, `action`, `formaction`, `sandbox`, `style`, `target`, and it deliberately re-emits `srcdoc` — so adding `"script"` via config yields `<iframe srcdoc="<script>…">` surviving sanitisation. Please either soften the docs wording or, better, filter the setting once in `config-reducer` so both consumers get the sanitised list — that would genuinely close the ticket in one place.

smohite-nice:
Thanks Dmitrii, all three points addressed in commit 94caffb5:

**Blocking 1 — type guard in sanitize.ts**
Added `Array.isArray()` check before the filter, plus `typeof tag === "string"` guard and `.trim()` on each element:
```ts
const list = Array.isArray(customAllowedHtmlTags) ? customAllowedHtmlTags : [];
const safeTags = list.filter(
    tag => typeof tag === "string" && !ALWAYS_BLOCKED_TAGS.has(tag.toLowerCase().trim()),
);
```

**Blocking 2 / Advisory — chat-components bypass**
Rather than only filtering in `sanitizeHTML`, the blocked-tags filter is now applied in `config-reducer.ts` at config-load time (both `SET_CONFIG` and `UPDATE_SETTINGS` cases), so the value stored in Redux is already clean. Since `@cognigy/chat-components` reads `customAllowedHtmlTags` from the same store key, it receives the pre-sanitised list automatically — the bypass is closed at the source.

`BLOCKED_TAGS` is defined inline in config-reducer to avoid the circular import (`sanitize.ts → store → config-reducer`). Both sets are identical; a comment flags that they must stay in sync.

**Docs — embedding.md**
Replaced "always blocked regardless of this setting" with "stripped from this list before it is applied — they cannot be re-enabled by tenant configuration", and added a note that non-string entries are silently ignored. This is accurate for both webchat and chat-components now that filtering happens in config-reducer.

**On FORBID_TAGS and the default path:** `FORBID_TAGS` in the base config is intentional defence-in-depth — if a future change inadvertently adds a dangerous tag back to `allowedHtmlTags`, DOMPurify still blocks it. The PR description has been updated to make clear this is an explicit hardening of the default path. Happy to coordinate merge order with #310 if you prefer the no-op approach.

kwinto (changes requested):
Re-reviewed after the 2026-09-11 commits — thanks, both of my earlier blocking items are resolved:

- **Malformed `customAllowedHtmlTags`** is now guarded with `Array.isArray()` + `typeof tag === "string"` + `.trim()`. I re-ran the four shapes (`["p", null]`, `["p", 1]`, `[{}]`, `"p,br"`) against the new filter — all handled gracefully, no `TypeError`.
- **`FORBID_TAGS` on the default path** is no longer a behaviour change: #310 is merged, and I verified none of the 12 tags remain in `allowedHtmlTags` on `main`, so it is a genuine no-op there. The success criterion "`customAllowedHtmlTags` not set → existing behaviour unchanged" now holds.
- **CSA-97603 closure** — filtering once in `config-reducer` is exactly the right place. `@cognigy/chat-components` reads the same `settings.widgetSettings.customAllowedHtmlTags`, so it now receives the sanitised list and the docs wording is accurate. I also checked `RESET_STATE` / `SET_PREV_STATE` in `reducer.ts` — they only restore `messages` and `rating`, so there is no path that reintroduces an unfiltered config.

All 11 checks are green. Two things left before I approve — both small and mechanical.

---

**1. (Blocking) The deny-list is duplicated in two files with only a comment holding it together.** `sanitize.ts:228-241`, `config-reducer.ts:7-21`

```ts
// Mirror of ALWAYS_BLOCKED_TAGS in sanitize.ts. Defined here to avoid a circular dependency
// (sanitize.ts imports storeRef which imports config-reducer). Both sets must stay in sync.
```

The circular-dependency reasoning is correct, but the fix for it is a shared constant, not a copy. A third file (`src/common/constants/blocked-tags.ts`) exporting the `Set` imports nothing from the store, so both consumers can import it with no cycle:

```ts
// src/common/constants/blocked-tags.ts
export const ALWAYS_BLOCKED_TAGS = new Set([...]);
```

`sanitize.ts` then re-exports it to preserve its current public export, and `config-reducer.ts` imports it directly. This matters because the drift it warns about is *already happening* — see point 2.

**2. (Blocking) `body`, `html`, `head`, `noframes` can still be re-enabled by tenant config.** `sanitize.ts:228-241`

#310 removed these four from the default allow-list and documented them as dangerous in the same comment block:

> `body / html / head` — structural document elements; no legitimate use in sanitised fragments
> `frame / frameset / noframes` — clickjacking and legacy frame injection

But they are not in `ALWAYS_BLOCKED_TAGS`, so `customAllowedHtmlTags: ["body", "noframes"]` puts them straight back. That is the exact scenario this PR exists to prevent, and it leaves the two PRs inconsistent: #310 says these tags are unsafe, #309 says a tenant may re-enable them. `frame` and `frameset` are already blocked here, so `noframes` being absent is clearly an oversight rather than a deliberate call.

Please add all four to the deny-list and to the `docs/embedding.md` row (16 tags instead of 12).

---

**Non-blocking, happy for these as follow-ups:**

- **8 of the 12 tags are still asserted vacuously.** `customAllowedHtmlTags.cy.ts:113-137` loops `expect(text).not.to.contain('<' + tag)` over all 12, but the payload only contains `script`, `iframe`, `form` and `object` markup — the assertions for `embed`, `applet`, `frame`, `frameset`, `meta`, `base`, `link` and `style` would pass against an empty implementation.
- **No test for the malformed shapes** (`[null]`, `[1]`, `"p,br"`) even though that was the whole point of `3017b8f4`, and no test covering the `config-reducer` filter — which is now the layer that protects chat-components.
- **The two layers disagree on the malformed-input fallback.** `config-reducer` maps a non-array to `undefined`, which falls back to the *full default allow-list* (permissive); `sanitize.ts` maps it to `[]`, which strips everything (restrictive). Worth aligning. Relatedly, `["script"]` now collapses to `[]`, so all HTML is stripped rather than just `script` — safe-fail and I think correct, but it deserves a sentence in `docs/embedding.md` since it is a behaviour change for anyone currently doing that.
- **PR description is stale.** "Changes" still lists only `sanitize.ts`; "Bot and agent messages are out of scope" is no longer accurate now that the reducer filter protects them (this is a good thing — it is what closes the ticket); and the Security section is ticked **"No security implications"** on an XSS-hardening PR, which will read oddly in an audit trail.

Please rebase on `main` before merge.

smohite-nice:
> Re-reviewed after the 2026-09-11 commits — thanks, both of my earlier blocking items are resolved:
> 
> * **Malformed `customAllowedHtmlTags`** is now guarded with `Array.isArray()` + `typeof tag === "string"` + `.trim()`. I re-ran the four shapes (`["p", null]`, `["p", 1]`, `[{}]`, `"p,br"`) against the new filter — all handled gracefully, no `TypeError`.
> * **`FORBID_TAGS` on the default path** is no longer a behaviour change: #310 is merged, and I verified none of the 12 tags remain in `allowedHtmlTags` on `main`, so it is a genuine no-op there. The success criterion "`customAllowedHtmlTags` not set → existing behaviour unchanged" now holds.
> * **[CSA-97603](https://nice-ce-cxone-prod.atlassian.net/browse/CSA-97603) closure** — filtering once in `config-reducer` is exactly the right place. `@cognigy/chat-components` reads the same `settings.widgetSettings.customAllowedHtmlTags`, so it now receives the sanitised list and the docs wording is accurate. I also checked `RESET_STATE` / `SET_PREV_STATE` in `reducer.ts` — they only restore `messages` and `rating`, so there is no path that reintroduces an unfiltered config.
> 
> All 11 checks are green. Two things left before I approve — both small and mechanical.
> 
> **1. (Blocking) The deny-list is duplicated in two files with only a comment holding it together.** `sanitize.ts:228-241`, `config-reducer.ts:7-21`
> 
> ```ts
> // Mirror of ALWAYS_BLOCKED_TAGS in sanitize.ts. Defined here to avoid a circular dependency
> // (sanitize.ts imports storeRef which imports config-reducer). Both sets must stay in sync.
> ```
> 
> The circular-dependency reasoning is correct, but the fix for it is a shared constant, not a copy. A third file (`src/common/constants/blocked-tags.ts`) exporting the `Set` imports nothing from the store, so both consumers can import it with no cycle:
> 
> ```ts
> // src/common/constants/blocked-tags.ts
> export const ALWAYS_BLOCKED_TAGS = new Set([...]);
> ```
> 
> `sanitize.ts` then re-exports it to preserve its current public export, and `config-reducer.ts` imports it directly. This matters because the drift it warns about is _already happening_ — see point 2.
> 
> **2. (Blocking) `body`, `html`, `head`, `noframes` can still be re-enabled by tenant config.** `sanitize.ts:228-241`
> 
> #310 removed these four from the default allow-list and documented them as dangerous in the same comment block:
> 
> > `body / html / head` — structural document elements; no legitimate use in sanitised fragments
> > `frame / frameset / noframes` — clickjacking and legacy frame injection
> 
> But they are not in `ALWAYS_BLOCKED_TAGS`, so `customAllowedHtmlTags: ["body", "noframes"]` puts them straight back. That is the exact scenario this PR exists to prevent, and it leaves the two PRs inconsistent: #310 says these tags are unsafe, #309 says a tenant may re-enable them. `frame` and `frameset` are already blocked here, so `noframes` being absent is clearly an oversight rather than a deliberate call.
> 
> Please add all four to the deny-list and to the `docs/embedding.md` row (16 tags instead of 12).
> 
> **Non-blocking, happy for these as follow-ups:**
> 
> * **8 of the 12 tags are still asserted vacuously.** `customAllowedHtmlTags.cy.ts:113-137` loops `expect(text).not.to.contain('<' + tag)` over all 12, but the payload only contains `script`, `iframe`, `form` and `object` markup — the assertions for `embed`, `applet`, `frame`, `frameset`, `meta`, `base`, `link` and `style` would pass against an empty implementation.
> * **No test for the malformed shapes** (`[null]`, `[1]`, `"p,br"`) even though that was the whole point of `3017b8f4`, and no test covering the `config-reducer` filter — which is now the layer that protects chat-components.
> * **The two layers disagree on the malformed-input fallback.** `config-reducer` maps a non-array to `undefined`, which falls back to the _full default allow-list_ (permissive); `sanitize.ts` maps it to `[]`, which strips everything (restrictive). Worth aligning. Relatedly, `["script"]` now collapses to `[]`, so all HTML is stripped rather than just `script` — safe-fail and I think correct, but it deserves a sentence in `docs/embedding.md` since it is a behaviour change for anyone currently doing that.
> * **PR description is stale.** "Changes" still lists only `sanitize.ts`; "Bot and agent messages are out of scope" is no longer accurate now that the reducer filter protects them (this is a good thing — it is what closes the ticket); and the Security section is ticked **"No security implications"** on an XSS-hardening PR, which will read oddly in an audit trail.
> 
> Please rebase on `main` before merge.

done

[CSA-97603]: https://nice-ce-cxone-prod.atlassian.net/browse/CSA-97603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kwinto (approved):

PR #312, opened 2026-08-26
fix(a11y): defer delete-confirm dialog focus so VoiceOver announces it (CGY-3274)

sushmi21:
## Ticket

[CGY-3274](https://nice-ce-cxone-prod.atlassian.net/browse/CGY-3274) — Delete Conversation Dialog, SC 4.1.2 Name, Role, Value: VoiceOver on macOS announces neither the dialog nor the focused element when the delete-confirmation dialog opens. NVDA is unaffected.

## Root cause

`DeleteConfirmModal` used React's `autoFocus` on the Cancel button, which fires `focus()` in the same task that inserts the dialog subtree into the DOM (the modal mounts via `CSSTransition mountOnEnter` at the moment it opens). VoiceOver processes the focus event against its not-yet-updated accessibility tree, so it announces neither the dialog role/title/description nor the focused Cancel button. NVDA re-resolves the focus context lazily, which is why the bug is VO-only.

## Fix

Route `DeleteConfirmModal` through `Modal`'s existing `initialFocusRef` path, which defers initial focus ~200ms so the accessibility tree settles first — the same mechanism already validated with VoiceOver for the disconnect overlay. Focus still lands on the safe default (Cancel); VO now announces dialog title → description → focused button.

This also activates `Modal`'s focus-restore-on-close for this dialog, which is safe: both consumers (delete conversation, delete all conversations) already restore focus to their trigger, and Modal's restore targets that same element behind a `document.contains` guard.

- `DeleteConfirmModal.tsx` — drop `autoFocus` (+ its eslint-disable), pass a Cancel ref as `initialFocusRef`
- `Modal.tsx` — `initialFocusRef` prop doc no longer recommends the autoFocus alternative
- `chatOptionsScreen.cy.ts` — updated comment; added focus-returns-to-trigger assertion after Esc (SC 2.4.3)
- `docs/accessibility.md` — card-variant guidance now mandates `initialFocusRef` and documents the VoiceOver race

## Validation

- `npm run lint:a11y` — 0 errors
- `tsc --noEmit` — no errors in touched files
- Cypress modal specs (`chatOptionsScreen.cy.ts`) run in CI; the existing `should("have.focus")` assertion tolerates the deferral since Cypress retries
- Manual VoiceOver-on-macOS verification recommended before merge

## How to test

**Setup:** `npm run dev` (http://localhost:8787) and embed the webchat with the delete option enabled:

```js
initWebchat("<endpoint-url>", {
	settings: {
		chatOptions: { enabled: true, enableDeleteConversation: true },
	},
});
```

**VoiceOver (macOS — the bug):**

1. Start VoiceOver (Cmd+F5), open the webchat, open the header menu (⋮) → chat options.
2. Activate the **Delete** button under "Delete conversation".
3. **Expected:** VoiceOver announces the dialog — "Delete conversation, dialog" — followed by the confirmation text ("Are you sure you want to delete this conversation? …") and then "Cancel, button" as focus lands (~200ms after open). Before this fix, focus moved silently and nothing was announced.
4. Press Esc — **expected:** the dialog closes and VO announces the "Delete" trigger button, which regains focus.
5. Same check via Previous Conversations → "Delete all conversations" (`previousConversations.enableDeleteAllConversations: true`).

**Regression (any browser, no screen reader):**

1. Open the same dialog — focus visibly lands on **Cancel** shortly after it opens.
2. Tab/Shift+Tab stay trapped inside the dialog (close X → Cancel → Delete anyway → wraps).
3. Esc and the X button close it and return focus to the trigger; "Delete anyway" still deletes the conversation.
4. **NVDA (Windows):** unchanged — dialog title, description, and Cancel are announced on open as before.

[CGY-3274]: https://nice-ce-cx-sandbox.atlassian.net/browse/CGY-3274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kwinto (approved):
Approving — the root-cause analysis is correct and the fix reuses the already-VoiceOver-validated `initialFocusRef` path rather than inventing a new mechanism.

Things I checked rather than assumed:

- **Ref forwarding works.** `CancelButton` is `styled(SecondaryButton)` → `styled(Button)` → `styled.button`; Emotion forwards `ref` through the whole chain, so it lands on the real DOM button.
- **The focus-restore branch actually runs.** Both consumers (`DeleteConversation`, `DeleteAllConversations`) keep `DeleteConfirmModal` permanently mounted and only toggle `isOpen`, so the effect's `else` branch fires on close — it would have been dead code had either consumer unmounted the modal.
- **The new Esc assertion will pass.** `DeleteConversation.onClose` focuses the trigger synchronously and `Modal` then restores to the same element, so `.webchat-delete-conversation-button` genuinely has focus.

Two minor notes, neither blocking:

1. **A ~200ms window where the focus trap is bypassable.** The Tab handler only intervenes when `document.activeElement` is the dialog's first/last focusable. Until the deferred `focus()` lands, activeElement is still the trigger — outside the dialog — so Tab moves natively into background content, and the `card` variant uses `<dialog open>` without `showModal()`, so nothing is inert. Narrow, but it's a small regression versus synchronous `autoFocus`. If you want to close it cheaply: focus the dialog container (`tabIndex={-1}`) synchronously on open, then move to `initialFocusRef` after the delay — that also gives VO a dialog boundary to announce.
2. **Focus restore now has two owners.** `DeleteConversation.onClose` focuses the trigger and `Modal` then does it again. Harmless today since both target the same element, but it means `Modal` will silently override any future consumer that wants focus elsewhere. Dropping the now-redundant consumer call would leave one owner.

Separately, `handleConfirmDelete` still ends with focus on `<body>` — it dispatches `setShowChatOptionsScreen(false)`, unmounting the trigger, so the `document.contains` guard correctly skips restore. Pre-existing and out of scope here, but it's the remaining SC 2.4.3 gap on the confirm path and probably deserves a follow-up ticket.

Agreed that manual VoiceOver verification should gate the merge — no automated test in this repo can prove the actual fix.

PR #313, opened 2026-08-27
CGY-3852 fix: retry reconnecting after a failed first connection attempt

vj-venkatesan:
# Success criteria

- A widget that goes offline before ever successfully connecting, then sends a message (e.g. via a Home Screen conversation-starter button), must still get that message delivered once the network comes back — the connection must actually be retried, not stuck forever.
- A widget that has never been opened/interacted with must still NOT attempt to reconnect merely on a network blip or tab-visibility change.
- Resuming a previous conversation (which connects the socket outside the normal redux `CONNECT` action) must keep the same auto-reconnect behavior it had before this change.

# How to test

1. Open the widget with `homeScreen.conversationStarters` enabled, go offline (devtools network throttling) before ever coming online, click a conversation-starter button to send a message, then go back online — the bot's reply should eventually arrive. See `cypress/e2e/reconnection.cy.ts` → "should send a Home Screen conversation-starter message after network reconnection".
2. Open the widget, go offline then online again without any interaction — no connection attempt should have been made (`store.getState().connection.hasAttemptedConnection` stays `false`). See the same file → "does not attempt to connect on a network blip if the widget was never opened".

# Root cause

`shouldReestablishConnection()` (`src/webchat/helper/connection-watchdog.ts`) gated the `NETWORK_ON`/`SET_PAGE_VISIBLE` reconnect triggers behind `autoInject.isConnectedOnce`, a flag that only becomes `true` after a *successful* connection. If the very first connection attempt happens while already offline, that attempt fails, the flag never flips, and nothing ever retries again — the buffered message is stuck forever. This PR replaces that gate with a new `connection.hasAttemptedConnection` flag, latched on either a `CONNECT` dispatch or a successful `SET_CONNECTED` (so it stays a strict superset of the old gate across every code path, including `switchSession`'s direct-connect flow used by "resume previous conversation").

# Security

- [ ] XSS

# Accessibility (WCAG 2.2 AA)

- [x] No accessibility implications (non-UI change — pure Redux state logic)

# Additional considerations

None.

🤖 Generated with a debugging/planning/multi-agent-implementation session using Claude Code.


kwinto (commented):
## Review — verified, not just read

I checked out `main`, cherry-picked this PR's spec onto it, built and ran Cypress locally, and read the `@cognigy/socket-client@5.0.0-beta.26` source. **The fix is correct and the root-cause analysis holds.** Approve with changes, mostly around the tests.

### What I could confirm

| Claim | Verdict |
|---|---|
| Root cause: no retry after a failed *first* connect | ✅ In `socket-client`, `setupReconnectInterval()` is only called from inside `socket.on("connect")`. If the first attempt fails, the retry machinery is never installed, so the message sits in `messageBuffer` forever. Diagnosis is right. |
| New gate is a **strict superset** of the old one | ✅ Old gate = `SET_CONNECTED(true)`; new = `CONNECT ∥ SET_CONNECTED(true)`. `switchSession` bypasses `CONNECT` but still emits `socket/connect` → `SET_CONNECTED(true)`. No path loses auto-reconnect. |
| Test 1 genuinely guards the regression | ✅ **Empirically verified**: with only `cypress/e2e/reconnection.cy.ts` applied on top of unpatched `main`, `should send a Home Screen conversation-starter message after network reconnection` **fails**; it passes on this branch. Real guard. |
| Red CI is caused by this PR | ❌ No. The Firefox job fails on `screenReaderLiveRegion.cy.ts` → *"stays silent when a persisted conversation is restored after a page reload"*, and the identical failure is present on #314. Pre-existing flake. All three reconnection tests pass in Chrome. |

Bonus improvement worth calling out in the description: a user who goes to the chat screen while offline and sends *nothing* also used to be stuck showing "disconnected" forever, and now recovers.

### Challenges

**1. Test 2 is near-tautological.** I ran it on `main` as well — it fails there, but only because `connection.hasAttemptedConnection` is `undefined`, so `should("be.false")` throws. It asserts the field this PR introduces, not observable behaviour. It would also pass for the wrong reason: with `homeScreen.enabled: true`, `Webchat._open()` never dispatches `showChatScreen()`, so no connect happens *regardless of the watchdog gate*. Please assert something behaviour-level (`connection.connected` / `connecting` false, or spy on `client.connect`), plus the pre-`goOffline()` assertion Copilot suggested.

**2. Test 2's title is wrong** — it says "never opened" but calls `cy.openWebchat()`. It actually tests "stayed on the Home Screen".

**3. Untested claim.** The description explicitly promises the `switchSession` / "resume previous conversation" path is preserved, but there is no test for it — and it's the riskiest path, since `switchSession` → `client.disconnect()` *clears* `socketReconnectInterval`.

**4. Module cycle smell.** `connection-reducer.ts` now imports from `connection-middleware.ts`, which imports back from the reducer. It's `import type`, so babel elides it — but drop the `type` keyword once and it becomes a genuine runtime cycle. Moving `CONNECT` / `connect()` into `connection-reducer.ts` next to the other action creators is ~5 lines and removes the cycle.

**5. Widened blast radius for permanently-broken endpoints.** Previously a never-successful connection was terminal. Now any widget that reached the chat screen retries on *every* `visibilitychange` and `online` event for the life of the page, and `shouldReestablishConnection()` does not consult `reconnectionLimit`. Each failed `client.connect()` also orphans a socket.io Manager (`this.socket` is only assigned in the `connect` handler; the reject path never closes the socket). Low severity, but `&& !reconnectionLimit` may be worth adding.

**6. Two overlapping latches now exist** — `autoInject.isConnectedOnce` and `connection.hasAttemptedConnection`. Also `hasAttemptedConnection` is set by `SET_CONNECTED(true)`, which is a *success*, not an *attempt*; `connectionRequested` or `shouldStayConnected` would read truer. A one-line comment noting it is a deliberate never-reset latch would help the next reader.

**7. Strategically this is a workaround.** The actual defect is in `socket-client` (`setupReconnectInterval` is reachable only on success). Fixing it upstream would fix every consumer. Fine to ship this now — worth an upstream ticket.

### Nits
- Reducer style inconsistency: `CONNECT` uses an early `return state` guard while `SET_CONNECTED` uses `state.x || action.connected`. Pick one.
- No unit-test runner in this repo (Cypress-only per `CLAUDE.md`), so the heavy E2E is justified — which makes point 1 matter more.

cypress/e2e/reconnection.cy.ts
```diff
@@ -103,6 +103,63 @@ describe("Reconnection", { browser: "!firefox" }, () => {
...
+
+		goOnline();
+		assertOnline();
+
+		cy.getWebchat()
+			.its("store")
+			.invoke("getState")
+			.its("connection.hasAttemptedConnection")
+			.should("be.false");
+	});
```
kwinto: I ran this test against unpatched `main` (spec cherry-picked, everything else untouched) and it **fails there** — but only because `connection.hasAttemptedConnection` is `undefined`, so `should("be.false")` throws. So it isn't guarding old behaviour; it's asserting the field this PR introduces.

It also passes for the wrong reason: with `homeScreen.enabled: true`, `Webchat._open()` skips `showChatScreen()` entirely (`src/webchat/components/Webchat.tsx`), so no `CONNECT` is dispatched **regardless of the watchdog gate**. And the title says "never opened" while the body calls `cy.openWebchat()`.

Suggested rewrite — renames it, locks in the scenario before going offline (Copilot's point), and adds behaviour-level assertions so it still means something if the flag is ever renamed:

```suggestion
	it("does not attempt to connect on a network blip while sitting on the Home Screen", () => {
		cy.initWebchat({
			settings: {
				homeScreen: { enabled: true },
			},
		});

		cy.openWebchat();

		const assertNeverConnected = () =>
			cy
				.getWebchat()
				.its("store")
				.invoke("getState")
				.then(state => {
					expect(state.connection.hasAttemptedConnection, "hasAttemptedConnection")
						.to.be.false;
					expect(state.connection.connected, "connected").to.be.false;
					expect(state.connection.connecting, "connecting").to.be.false;
				});

		// Lock in the scenario: staying on the Home Screen must not have
		// triggered a connection attempt, otherwise the offline/online cycle
		// below would no longer exercise the watchdog gate at all.
		assertNeverConnected();

		goOffline();
		assertOffline();

		goOnline();
		assertOnline();

		assertNeverConnected();
	});
```

src/webchat/store/connection/connection-reducer.ts
```diff
@@ -1,15 +1,18 @@
 import { Reducer } from "redux";
+import type { ConnectAction } from "./connection-middleware";
```
kwinto: This closes a module cycle: `connection-reducer` → `connection-middleware` → `connection-reducer` (the middleware imports `setConnecting` / `setReconnectionLimit` from here).

It is harmless *today* because `import type` is erased and the repo compiles with `babel-loader` (per-file transpile). But it is a trap — delete the `type` keyword during a future refactor and it silently becomes a real runtime cycle, with `connect` possibly `undefined` at module-init depending on evaluation order.

Cheapest fix: move the action to the reducer, where every other connection action creator already lives.

```ts
// connection-reducer.ts
export const CONNECT = "CONNECT";
export const connect = () => ({ type: CONNECT as "CONNECT" });
export type ConnectAction = ReturnType<typeof connect>;
```

```ts
// connection-middleware.ts — drop the local CONNECT/connect/ConnectAction
import { connect, setConnecting, setReconnectionLimit } from "./connection-reducer";
export type { ConnectAction } from "./connection-reducer"; // if external callers still import it from here
```

Dependency then points one way only (middleware → reducer), which is the usual Redux layering.

src/webchat/helper/connection-watchdog.ts
```diff
@@ -1,7 +1,7 @@
 import { StoreState } from "../store/store";
 
 export const shouldReestablishConnection = (store: StoreState) =>
-	store?.autoInject?.isConnectedOnce;
+	store?.connection?.hasAttemptedConnection;
```
kwinto, line 4: The behavioural trade-off here deserves a conscious decision, because the blast radius widens.

Before, a connection that **never** succeeded was terminal: the gate required `SET_CONNECTED(true)`. Now any widget that merely reached the chat screen (`SHOW_CHAT_SCREEN` → `CONNECT`) will re-attempt on *every* `visibilitychange` and *every* `online` event, for the life of the page — including endpoints that are permanently misconfigured or down. `reconnectionLimit` (the flag that drives the permanent "we gave up" overlay) is not consulted at all.

Each failed attempt also orphans a socket.io Manager inside `socket-client`: `this.socket` is only assigned in the `socket.on("connect")` handler, and the `connect_error` reject path never closes the socket it created. Small per attempt, but it now repeats on every tab focus.

Worth considering:

```ts
export const shouldReestablishConnection = (store: StoreState) =>
	!!store?.connection?.hasAttemptedConnection && !store?.connection?.reconnectionLimit;
```

Note this would be a behaviour change for the *already-connected* case too (which currently does retry past the limit), so it needs a product call — but as written, "gave up permanently" and "retry on every focus" contradict each other.

For context, the underlying defect is upstream: `setupReconnectInterval()` in `@cognigy/socket-client@5.0.0-beta.26` is only reachable from inside `socket.on("connect")`, so a failed *first* attempt leaves no retry machinery at all. Fixing it there would fix every consumer; this PR is a reasonable Webchat-side workaround in the meantime, but an upstream ticket would be good.

src/webchat/store/connection/connection-reducer.ts
```diff
@@ -60,6 +64,15 @@ export const connection: Reducer<
...
 		}
 
+		case "CONNECT": {
+			if (state.hasAttemptedConnection) return state;
+
+			return {
+				...state,
+				hasAttemptedConnection: true,
+			};
+		}
```
kwinto: **The fix is incomplete: the same bug survives on the "resume previous conversation" path.** I verified this by running a probe spec against this branch.

The PR description says the latch "stays a strict superset of the old gate across every code path, including `switchSession`'s direct-connect flow". That is true — it *is* a superset, so nothing regresses. But superset-of-a-broken-gate is still broken: on the resume path the latch is armed **only if the connection succeeds**, which is precisely the condition this PR set out to stop depending on.

Why `SWITCH_SESSION` never arms it:

- Resuming dispatches only `switchSession(...)` + `setShowPrevConversations(false)` (`ConversationsList.tsx` → `ConnectedWebchatUI.tsx`). No `showChatScreen()`, so **no `CONNECT`**.
- `createPrevConversationsMiddleware` calls `client.switchSession()` directly, which does `disconnect()` (clearing `socketReconnectInterval`) then `connect()`. If that rejects, the `.catch()` only does `setConnecting(false)`.
- No `CONNECT`, no `SET_CONNECTED(true)` ⇒ `hasAttemptedConnection` stays `false` ⇒ `NETWORK_ON` / `SET_PAGE_VISIBLE` remain gated off forever.

Probe result on this branch (mock endpoint, socket never completes):

```
✓ A: SWITCH_SESSION alone does not arm hasAttemptedConnection (2628ms)
✓ B: a successful connect does arm the latch (control) (587ms)
```

User-visible consequence — the mirror image of the bug being fixed:

1. Open widget → **Previous conversations** → click a conversation, while offline.
2. `switchSession`'s connect rejects; the reconnect interval was just cleared by `disconnect()`.
3. Network returns → `NETWORK_ON` is gated off → the widget stays disconnected indefinitely.

It self-heals *if* the user types (`SEND_MESSAGE` → `dispatch(connect())` arms the latch), so the buffered-send story is fine — but passive recovery is not, and the connection-lost overlay sticks around.

Three-line fix, consistent with the intent of the PR ("we requested a connection" is the real trigger, and `SWITCH_SESSION` *is* a connection request):

```suggestion
		case "CONNECT":
		// A session switch connects the socket directly, bypassing CONNECT.
		// It is still a deliberate connection request, so it must arm the
		// reconnect latch even if this first attempt fails (e.g. offline).
		case "SWITCH_SESSION": {
			if (state.hasAttemptedConnection) return state;

			return {
				...state,
				hasAttemptedConnection: true,
			};
		}
```

(`SwitchSessionAction` would join the reducer's action union — importing it from `../previous-conversations/previous-conversations-middleware` has the same cycle caveat I noted on line 2, so it is one more reason to keep action creators in reducers.)

This also removes the last behavioural difference between the two entry points, which would make the "strict superset" claim in the description an actual equivalence, and it gives challenge #3 from my summary review a home: a test for the resume path.

kwinto:
### Follow-up: I verified challenge #3, and it turned into a real gap

I said the `switchSession` / "resume previous conversation" claim was untested. I built this branch and ran a probe spec against it. Result: **the fix is incomplete on that path** — details and a suggested patch in [this comment](https://github.com/Cognigy/Webchat/pull/313#discussion_r3880124481).

Short version: resuming a conversation dispatches only `SWITCH_SESSION`, never `CONNECT`, and `client.switchSession()` connects directly. So the latch is armed **only when the connection succeeds** — exactly the dependency this PR exists to remove. Offline-resume → network returns → still stuck.

```
✓ A: SWITCH_SESSION alone does not arm hasAttemptedConnection (2628ms)
✓ B: a successful connect does arm the latch (control) (587ms)
```

To be fair to the change: this is **not a regression**. The old `isConnectedOnce` gate had the same hole, so the "strict superset" claim in the description is accurate. It is just that the superset is still missing the case the PR is about, on the second of the two entry points. Adding `case "SWITCH_SESSION"` to the same latch closes it and makes the description's claim an equivalence rather than a superset.

Everything else from my review stands, and the core fix + test 1 are solid — test 1 demonstrably fails on `main` and passes here.

kwinto:
### Checked the sibling PR — it does **not** cover the gap I raised

For anyone following both: [Cognigy/SocketClient#77](https://github.com/Cognigy/SocketClient/pull/77) is the same ticket (CGY-3852) by the same author, so I reviewed it too and checked whether it removes the need for point 7 of my review (*"the real defect is upstream"*). It does not — the two PRs fix **different** failure modes:

| | Failure mode fixed |
|---|---|
| SocketClient#77 | Message lost **after** a successful send, in the disconnect-before-ack window |
| This PR (#313) | **No reconnect ever attempted** when the *first* connection attempt fails |

I built SocketClient#77 and confirmed `setupReconnectInterval()` is untouched — still only reachable from inside `socket.on("connect")` (`src/socket-client.ts:352`). So a first `connect()` that rejects still installs no retry interval at all in **every** consumer of the client, not just Webchat. This PR's Redux-side latch remains necessary, and the upstream fix is still outstanding. Worth a follow-up ticket.

One thing worth coordinating, though: SocketClient#77 adds `this.messageBuffer = []` to `switchSession()`. I probed it and a message the user typed **while offline** is now silently discarded when a session switch happens:

```
PROBE P1 delivered = []
✓ P1: a message buffered while offline is SILENTLY DROPPED by switchSession
```

Webchat reaches that path from "Start new conversation" and "Delete conversation" (both dispatch `switchSession()` alone). It does **not** affect this PR's test 1, because `handleSendActionButtonMessage` dispatches `onSwitchSession()` *before* `onSendMessage()`, so the buffer clear happens first — but the two changes do interact, and once #77 lands it's worth a Webchat-side test pinning that ordering, since silently reordering those two dispatches would start dropping conversation-starter messages.

kwinto:
**Follow-up: I implemented and validated the `SWITCH_SESSION` fix from [my earlier comment](https://github.com/Cognigy/Webchat/pull/313#discussion_r3880124481).**

Rather than leave that as an assertion, I applied it to this branch, built, and ran it both ways. The test discriminates cleanly:

| | `arms the latch when a previous conversation is resumed` |
|---|---|
| This branch as-is | ✗ `AssertionError: expected false to be true` |
| With the 4-line patch | ✓ passes (2030ms) |

And the guard case — *"still does not arm the latch by merely sitting on the Home Screen"* — passes in **both** runs, so the patch does not widen the gate for the "never interacted" scenario your test 2 is protecting.

The patch (`connection-reducer.ts`):

```ts
import type { SwitchSessionAction } from "../previous-conversations/previous-conversations-middleware";

export const connection: Reducer<
	ConnectionState,
	| SetConnectedAction
	| SetReconnectionLimitAction
	| SetConnectingAction
	| ConnectAction
	| SwitchSessionAction
> = (state = initialState, action) => {
	switch (action.type) {
		// …

		case "CONNECT":
		// A session switch connects the socket directly, bypassing CONNECT.
		// It is still a deliberate connection request, so it must arm the
		// reconnect latch even if this first attempt fails (e.g. offline).
		case "SWITCH_SESSION": {
			if (state.hasAttemptedConnection) return state;

			return {
				...state,
				hasAttemptedConnection: true,
			};
		}
```

Two notes:

1. This adds a **second** reducer→middleware `import type`, which reinforces the point on line 2 — with `CONNECT` and `SWITCH_SESSION` both defined in middleware files, the reducer now type-depends on two modules that depend back on it. Moving both action creators into their reducers removes the cycle *and* both `import type` lines.
2. It also makes the description's "strict superset" claim an actual **equivalence** — after this, every path that requests a connection arms the latch, whether or not the attempt succeeds, which is exactly the invariant the PR is reaching for.

Happy to push this as a commit to your branch if that's easier than re-typing it.

src/webchat/store/connection/connection-reducer.ts
vj-venkatesan: Applied - added `case "SWITCH_SESSION"` to the same latch in `5002e9cc`, matching your patch. Verified it the same way: reverted the case locally, confirmed the new Cypress test ("arms the latch when a previous conversation is resumed") fails, restored it and confirmed it passes, and that the existing "never interacted" guard test still passes. Also moved `SWITCH_SESSION`/`switchSession` into `previous-conversations-reducer.ts` (mirroring the `CONNECT` move on the other thread) so the reducer's import of `SwitchSessionAction` doesn't reintroduce the cycle.

src/webchat/store/connection/connection-reducer.ts
vj-venkatesan: Fixed in 5002e9cc - moved `CONNECT`/`connect()` into `connection-reducer.ts` and `SWITCH_SESSION`/`switchSession()` into `previous-conversations-reducer.ts`. Both action creators now live in their reducers; the middleware files import from the reducers instead of the reverse. Updated all call sites (`Webchat.tsx`, `ConnectedWebchatUI.tsx`, `DeleteConversation.tsx`, `DeleteAllConversations.tsx`, `analytics-middleware.ts`) accordingly.

cypress/e2e/reconnection.cy.ts
vj-venkatesan: Applied your suggested rewrite verbatim in 5002e9cc - renamed the test, added `assertNeverConnected()` before `goOffline()` to lock in the scenario, and made it assert `connected`/`connecting` alongside `hasAttemptedConnection`.

cypress/e2e/reconnection.cy.ts
```diff
@@ -103,6 +103,63 @@ describe("Reconnection", { browser: "!firefox" }, () => {
...
+
+	it("does not attempt to connect on a network blip if the widget was never opened", () => {
+		cy.initWebchat({
+			settings: {
+				homeScreen: { enabled: true },
+			},
+		});
+
+		cy.openWebchat();
+
```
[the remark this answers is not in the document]
vj-venkatesan: Fixed in 5002e9cc - renamed to "does not attempt to connect on a network blip while sitting on the Home Screen" (via kwinto's suggested rewrite, same thread).

cypress/e2e/reconnection.cy.ts
```diff
@@ -103,6 +103,63 @@ describe("Reconnection", { browser: "!firefox" }, () => {
...
+					enableConnectionStatusIndicator: false,
+				},
+			},
+		});
+
+		// Open Webchat - stay on the Home Screen, never connect while online.
+		cy.openWebchat();
+
+		// Go offline before the very first connection attempt.
+		goOffline();
```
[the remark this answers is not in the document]
vj-venkatesan: Fixed in 5002e9cc - now asserts `hasAttemptedConnection`/`connected`/`connecting` are all false before `goOffline()` as well as after the online/offline cycle (via kwinto's suggested rewrite, same thread).

src/webchat/helper/connection-watchdog.ts
vj-venkatesan: Agreed this needs a product call, not a pure bugfix - leaving `shouldReestablishConnection()` as-is for this PR rather than guessing at the intended behavior for the already-connected-then-dropped case. Filed CGY-37178 to track the upstream `setupReconnectInterval()` root cause you identified (which is the more fundamental fix - every consumer gets no retry machinery at all after a failed first connect), and will raise the `reconnectionLimit` gating question separately as a product decision rather than bundle it here.

vj-venkatesan:
Pushed 5002e9cc addressing the review feedback:

- **The `SWITCH_SESSION` gap (confirmed real, not just theoretical)**: added `case "SWITCH_SESSION"` to the same latch alongside `CONNECT`, per kwinto's verified patch. Confirmed the regression test fails without it and passes with it.
- **Test quality**: applied kwinto's suggested rewrite of the "never opened" test (renamed, locks in the pre-offline state, asserts `connected`/`connecting` too) - addresses both Copilot comments on that test.
- **New regression test**: added a Cypress test for the resume-path fix ("arms the latch when a previous conversation is resumed").
- **Module cycle**: moved `CONNECT`/`connect()` into `connection-reducer.ts` and `SWITCH_SESSION`/`switchSession()` into `previous-conversations-reducer.ts`, so both action creators live with their reducers instead of the reducer importing (even type-only) from a middleware file that imports back from it.
- **Upstream root cause**: filed [CGY-37178](https://cognigy.atlassian.net/browse/CGY-37178) against `@cognigy/socket-client` for `setupReconnectInterval()` never running after a failed first `connect()`, linked to CGY-3852.

Deliberately **not** changed: `shouldReestablishConnection()` gating on `reconnectionLimit` - as kwinto noted, that's a behavior change for the already-connected-then-dropped case too and needs a product decision, not a drive-by fix in this PR.

All 19 specs in `reconnection.cy.ts` pass, plus `prevConvesations.cy.ts`, `homeScreen.cy.ts`, and `closeButtonAnalytics.cy.ts` as a broader smoke check on the moved modules. `lint:a11y` is clean.

[CGY-37178]: https://nice-ce-cxone-prod.atlassian.net/browse/CGY-37178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

kwinto:
Re-reviewed at `7abd382`. All five points are addressed, and I re-verified each one independently rather than taking the replies at face value:

| Finding | Status | How I checked |
|---|---|---|
| Incomplete fix on the resume path | ✅ Fixed | `connection-reducer.ts:79-80` now falls `case "CONNECT": case "SWITCH_SESSION":` into the same latch |
| Module cycle | ✅ Genuinely gone | `connect`/`CONNECT` moved to `connection-reducer`, `switchSession`/`SWITCH_SESSION` to `previous-conversations-reducer`. Confirmed `previous-conversations-reducer` has **no** import back into `connection-reducer`, so the new reducer→reducer edge is a DAG, not a relabelled cycle. The remaining `import type { SwitchSessionAction }` is one-directional. |
| Near-tautological test 2 | ✅ Fixed | Renamed, and `assertNeverConnected()` now pins `connected`/`connecting` before *and* after the offline/online cycle, so it fails loudly if `openWebchat()` ever starts connecting |
| Watchdog blast radius | ✅ Correctly deferred | Agreed this is a product call, not a bugfix. Leaving `shouldReestablishConnection()` alone is the right conservative choice here. |
| Upstream `setupReconnectInterval()` | ✅ Tracked | CGY-37178 |

Two things I checked that nobody raised:

**The new test is faithful, not just green.** It dispatches a raw `{ type: "SWITCH_SESSION", sessionId }` rather than calling the `switchSession()` creator. That's normally a smell — a hand-rolled literal can drift from the real action shape and leave the test passing while production breaks. Here it's safe: the creator emits exactly `type: "SWITCH_SESSION"` with `conversation` optional, so the literal is shape-compatible, and any future rename of the constant would make this test *fail* rather than silently pass. No change needed; noting it so the next reader doesn't have to re-derive it.

**CI is fully green, including the Firefox job.** That job was failing on `screenReaderLiveRegion.cy.ts` earlier in this PR's life. I'd already established it was a pre-existing flake (it reproduced identically on #314), and it now passes at this head — so it was never related to this change.

No further objections from me. Not marking this approved since that's the maintainers' merge gate to hold, but from a correctness standpoint I have nothing left outstanding.

One residual, deliberately not blocking: `WebchatUI.tsx:1116` (`handleSendActionButtonMessage`) still relies on `onSwitchSession()` being dispatched immediately before `onSendMessage()`. That ordering is load-bearing but unenforced. It's untouched by this PR and safe today — worth a comment there at some point so a future reorder doesn't quietly break it.

kwinto (approved):
Approve.

The `connection.hasAttemptedConnection` latch is the right shape for this bug — gating the watchdog on "we tried" rather than "we previously succeeded" is what makes the cold-start failure recoverable, and it covers the Home Screen conversation-starter and resumed-conversation paths too.

Earlier feedback on the resume path and test quality looks addressed on the current head. Cypress coverage for the offline-first and no-interaction cases is a good addition, CI is 9/9, and the UI file changes are import-only so there is no a11y surface here.

No blocking issues.

PR #315, opened 2026-08-27
CGY-36979 test: control run for the E2E Firefox failure (to be closed)

ls-ramos:
Temporary control run, not for review or merge.

This branch carries only the audit job from #314, with no dependency changes, to check whether `Progressive Rendering with Plugin Messages > should immediately show plugin messages without waiting for animation queue` fails on Firefox independently of that pull request. The audit check itself is expected to fail here, since the lock file on `main` still has the findings that #314 fixes.

It will be closed as soon as the Firefox run is read.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,6 +3,20 @@ name: Lint
 on: [pull_request]
 
 jobs:
+    # Dependency vulnerability gate for pull requests (CGY-36979). Audits
+    # package-lock.json, so no install is needed. Limited to production dependencies,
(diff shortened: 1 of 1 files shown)


ls-ramos:
Control run read: E2E Firefox fails on this branch too, with no dependency changes at all, so the failure on #314 is independent of it. Interesting detail, this control failed on a different spec (Screen Reader Live Region) and also failed E2E Progressive Rendering, which passes on #314, so the fragility moves around between runs rather than sitting in one test.

The audit check failing here is the expected result and a useful confirmation, it is reading the lock file on main which still has the findings that #314 fixes.

Closing, thank you!

